Issue #1, opened 2016-12-20
Signature's native JSON marshalling does not encode instances into base64 format


alessio:
`Signature` concrete types should define their own `MarshalJSON()` and `UnmarshalJSON()` methods as currently marshalling objects does not work as expected. In particular, `json.Marshal()` returns a stringified array of integers; byte arrays should be marshalled into base64 strings instead. The following simple program may be used to reproduce the issue:
```go
package main

import (
	"encoding/json"
	"fmt"

	"github.com/tendermint/go-crypto"
)

func main() {
	var sig1 crypto.Signature

	privKey := crypto.GenPrivKeyEd25519()
	sig1 = privKey.Sign([]byte("sample string"))
	jsonBytes := mustMarshal(sig1)
	fmt.Printf("%q\n", string(jsonBytes))
}

func mustMarshal(v interface{}) []byte {
	ret, err := json.Marshal(v)
	if err != nil {
		panic(err)
	}
	return ret
}
```

alessio:
Please note that the same issue affects `crypto.PrivKey` and `crypto.PubKey` concrete types too.

ramilexe:
>byte arrays should be marshalled into base64 strings instead

why? doesn't stringified array of integers represent byte array?

alessio:
@ramilexe, it might, though it's weird and the JSON marshalling output is inconsistent:
```shell
alessio@bizet:~/devel/crypto$ cat sample.go ; go run sample.go 
package main


import (
	"encoding/json"
	"fmt"

	"github.com/tendermint/go-crypto"
)

func main() {
	privKeySecp256k1 := crypto.GenPrivKeySecp256k1()
	privKeyEd25519 := crypto.GenPrivKeyEd25519()

	// SignatureEd25519
	origEd25519 := privKeyEd25519.Sign([]byte("sample string"))
	fmt.Printf("privKeyEd25519:\n %q\n", string(mustMarshal(privKeyEd25519)))
	fmt.Printf("privKeyEd25519.PubKey():\n %q\n", string(mustMarshal(privKeyEd25519.PubKey())))
	fmt.Printf("SignatureEd25519:\n %q\n", string(mustMarshal(origEd25519)))

	// SignatureSecp256k1
	origSecp256k1 := privKeySecp256k1.Sign([]byte("sample string"))
	fmt.Printf("privKeySecp256k1:\n %q\n", string(mustMarshal(privKeySecp256k1)))
[...]

PR #2, opened 2016-12-20
Implement signatures MarshalJSON()

alessio:
Cast signatures to byte slices before calling
json.Marshal().

Fixes https://github.com/tendermint/go-crypto/issues/1


ebuchman:
@alessio and @ethanfrey we can close this now?

alessio:
Yes, we can. Closing now, thanks!

PR #3, opened 2017-02-23
Use go-data for nice json un/marshaling

ethanfrey:
Another approach to solve #1 

This uses wrapping structs to allow json unmarshalling of interfaces with minimal boilerplate code (which I hope to auto-generate one day). The default behavior with wire.BinaryBytes() and wire.JSONBytes() is not changed at all.

If you use the `encoding/json` library, we get nice handling of the three interfaces, `PubKey`, `PrivKey` and `Signature`.  You must just store them in a wrapping struct: `PubKeyS`, `PrivKeyS` or `SignatureS`.  Please take a look at the tests.

We use `go-data` to export all byte slices (and with custom JSONMarshal also the byte arrays) as hex strings.  However this is flexible, and you can set the byte encoding to use hex, base64 or base58 in the main entry of your program and it will affect the encoding of all types that make use of `data.Bytes` or `data.Encoder`.

Example default encodings are for public keys are:

```
{
	"type": "ed25519",
	"data": "345347cee043dc6d6f70b0ae9a3fc5a8c977006fd13000680b1e2afecd0475d5"
}

{
	"type": "secp256k1",
	"data": "989f04749bcc689f6dec7489cc227bc3e2fbc4c37067df8750e0791d2e6297c106e244fc635ecd1696698dfd8d49efa752bb3ad7ff8c7fd23b79743cce8476b8"
}
```

If you use `crypto.PubKeyS` in your structs, you can cleanly Unmarshal this json into usable PubKey's as well.

### DIFF
--- a/signature.go
+++ b/signature.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	. "github.com/tendermint/go-common"
+	data "github.com/tendermint/go-data"
 	"github.com/tendermint/go-wire"
 )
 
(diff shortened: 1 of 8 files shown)


alessio:
:+1

signature.go
```diff
@@ -20,14 +21,39 @@ type Signature interface {
 const (
 	SignatureTypeEd25519   = byte(0x01)
 	SignatureTypeSecp256k1 = byte(0x02)
+	SignatureNameEd25519   = "ed25519"
```
ebuchman: what about just one `NameEd25519` instead of duplicating names for each type?

ebuchman (approved):
Very cool

ethanfrey:
Okay, I unified all the types and names with one variable per algorithm in 0e92dd5

I think this is what you wanted with your comment bucky.  If you like it, me too.  If not, I can just revert the change.

PR #6, opened 2017-03-28
Feature/improve godata

ebuchman:
Openning PR for review

### DIFF
--- a/embed_test.go
+++ b/embed_test.go
@@ -9,50 +9,13 @@ import (
 	data "github.com/tendermint/go-data"
 )
 
-type Foo struct {
-	Name string
-}
-
-func (f Foo) Greet() string {
-	return "Foo: " + f.Name
-}
-
-type Bar struct {
-	Age int
-}
-
-func (b Bar) Greet() string {
-	return fmt.Sprintf("Bar #%d", b.Age)
+type PubName struct {
+	PubNameInner
 }
 
 type PubNameInner interface {
-	Greet() string
-}
-
-type privNameInner interface {
-	Greet() string
-}
-
-type Greeter interface {
-	Greet() string
-}
-
-var (
-	pubNameMapper, privNameMapper data.Mapper
-)
-
-// register both public key types with go-data (and thus go-wire)
-func init() {
-	pubNameMapper = data.NewMapper(PubName{}).
-		RegisterImplementation(Foo{}, "foo", 1).
-		RegisterImplementation(Bar{}, "bar", 2)
-	privNameMapper = data.NewMapper(PrivName{}).
-		RegisterImplementation(Foo{}, "foo", 1).
-		RegisterImplementation(Bar{}, "bar", 2)
-}
-
-type PubName struct {
-	PubNameInner
+	AssertIsPubNameInner()
+	String() string
 }
 
 func (p PubName) MarshalJSON() ([]byte, error) {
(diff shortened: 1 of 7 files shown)


encode_test.go
```diff
@@ -95,17 +115,17 @@ func toFromJSON(t *testing.T, in interface{}, recvr interface{}) {
 
 func TestNilEncodings(t *testing.T) {
 	// make sure sigs are okay with nil
-	a, b := SignatureS{}, SignatureS{}
+	a, b := Signature{}, Signature{}
```
ebuchman: `var a, b Signature`

ebuchman (commented):
Looks good. I might stress a little more that the XXXInner interfaces are not to be used ....

embed_test.go
```diff
@@ -0,0 +1,125 @@
...
+	// assert := assert.New(t)
+	// require := require.New(t)
+
+	cases := []struct {
+		in, out  Greeter
+		expected string
+	}{
+		{PubName{Foo{"pub-foo"}}, &PubName{}, "Foo: pub-foo"},
+		{PubName{Bar{7}}, &PubName{}, "Bar #7"},
+		// Note these fail - if you can figure a solution here, I'll buy you a beer :)
```
ebuchman: don't these fail because the interface is private and hence can't be seen by wire ?

encode_test.go
ethanfrey: Yes, that is more idiomatic. Thanks

How to stress do not use except by changing the names? I'm open to suggestions

ebuchman: Comments are fine:

`DO NOT USE this interface. It is public by necessity but should never be used directly outside of this package. Only use the PubKey, never the PubKeyInner`

embed_test.go
ethanfrey: Exactly. I wrote the tests to verify the behavior before coverting all code.

ebuchman: Ok. So there's no real beer offering here and we can remove these ? Or do you want to assert in a separate test that they'll fail?

ethanfrey: I'd love a trick to make them private without breaking go-wire. Thus, if you can fix these tests, congrats!

ebuchman: Wouldn't that be breaking the type system? There should be no way to get an external package to read private stuff - though I'm not sure what happens if the field is public but the type is private

ethanfrey: I don't mean breaking the type system. But I got to use private fields in json by implementing custom Marshalers. I guess there is no equivalent in go- wire?
Like the fields are private, but there are public methods that modify them.

encode_test.go
ethanfrey: Awesome. Will do tomorrow

ethanfrey:
I think I fixed the issues.

The failing test is due to it requiring the develop branch of go-data (which has the breaking change of renaming RegisterInterface to RegisterImplementation).  And go-crypto has no glide file or `make get_vendor_deps`.  Should I add that???

ethanfrey:
The go-wire struct parsing seems to be dealt with [here](https://github.com/tendermint/go-wire/blob/master/reflect.go#L314-L318) and [here](https://github.com/tendermint/go-wire/blob/master/reflect.go#L524-L528)

They make use of `reflect.Value.Field` and the `TypeInfo` list that was created [here](https://github.com/tendermint/go-wire/blob/master/reflect.go#L148-L182)

Assuming the go-wire code were modified, one could add a simple interface to allow custom un/marshaling of go-wire binary, like:

```Go
type Reflector interface {
  NumFields() int
  Field(int) interface{}  // returning a pointer to the value at that location
}
```

Probably totally out of scope and unnecessary, but I write it here for what it is worth.  If you think it is useful, I can make an issue in go-wire...

PR #8, opened 2017-04-10
Reorg algorithms

ethanfrey:
I pulled ed25519 and secp256k1 implementations of the `PrivKey`, `PubKey` and `Signature` primitives into their own files.  This makes no changes in the API, but lays the ground-work for adding more algorithms, as the generic structs are in eg. `pub_key.go`, while all algorithm-dependent code is in eg. `ed25519.go`.

I could conceive of another file here (eg. `rsa.go`) or even sub-packages for each algorithm, and in the `main.go` file, one would import the different algorithms one wanted their app to support. (So if you only want secp256k1, it would be easy to compile without ed25519 support).

That is longer-range, for now, more for clarity and defining what exactly needs to be implemented by a new crypto algorithm.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,7 @@
 *.swp
 *.swo
+vendor
+
+# these should be locked in the including repo, the dependencies are mainly
+# for testing and documentation
+glide.lock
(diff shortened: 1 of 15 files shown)


ethanfrey:
Never mind... Aborting this attempt, maybe try something similar later.

Issue #14, opened 2017-05-12
Integrate mneumonics from mintkeys


ethanfrey:
Note there's also https://github.com/tendermint/mintkey for wordlist stuff

We should figure out some way to integrate this for exporting secret keys or verifying public keys

ethanfrey:
Fixed in #19

Issue #17, opened 2017-06-18
Better support for non-interactive password


ethanfrey:
There was a request for non interactive password entering for test scripts

https://github.com/tendermint/light-client/issues/23

This works with echo and pipe, but is the prompt is still output to stdout and needs to be trimmed from the json output 

We could support some extra fd as input as mentioned on the linked pages from the above issue... 

Like `--password=3` for a custom fd or `--password =-` to read from stdin without a prompt

ethanfrey:
Fixed in #19

PR #19, opened 2017-06-19
cli improvements

ethanfrey:
Various enhancements to the cli.

* Non-interactive passwords #17 
* Better cli tests (shutil)
* Support mneumonic backup / recovery #14


ethanfrey:
Improved `new` command, as well as adding `delete` and `recover` to resolve issue #14 

Note, that if you make an ed25519 key, the seed is *super* long. If you do`keys new demouser -t secp256k1` it is a much more reasonable length.

This has to do with us storing private keys of ed25519 as 64 bytes, when we only really need 32 bytes.... half of it is the pubkey. Do we need to store that?

https://github.com/tendermint/go-crypto/blob/develop/priv_key.go#L99-L104

cmd/delete.go
```diff
@@ -0,0 +1,49 @@
...
+	"fmt"
+
+	"github.com/pkg/errors"
+
+	"github.com/spf13/cobra"
+)
+
+// deleteCmd represents the delete command
+var deleteCmd = &cobra.Command{
+	Use:   "delete <name>",
```
rigelrozanski: Use square bracket `delete [name]` to be consistent with cobra, There are a number of instances of `<name>` to change

cmd/get.go
```diff
@@ -25,20 +25,18 @@ var getCmd = &cobra.Command{
 	Use:   "get <name>",
```
rigelrozanski: `[name]`

keys/cryptostore/holder.go
```diff
@@ -32,13 +39,35 @@ func (s Manager) assertKeyManager() keys.Manager {
...
 	err = s.es.Put(name, passphrase, key)
+	if err != nil {
+		return keys.Info{}, "", err
+	}
+	seed, err := s.codec.BytesToWords(key.Bytes())
+	phrase := strings.Join(seed, " ")
+	return info(name, key), phrase, err
+}
+
+func (s Manager) Recover(name, passphrase, seedphrase string) (keys.Info, error) {
```
rigelrozanski: Add Lint comments throughout! All exposed functions should have a header comment.

keys/cryptostore/holder.go
```diff
@@ -32,13 +39,35 @@ func (s Manager) assertKeyManager() keys.Manager {
...
+	if err != nil {
+		return keys.Info{}, err
+	}
+
+	key, err := crypto.PrivKeyFromBytes(data)
+	if err != nil {
+		return keys.Info{}, err
+	}
+
+	// d00d, it worked!  create the bugger....
```
rigelrozanski, line 74: :smile:

keys/wordcodec.go
```diff
@@ -0,0 +1,199 @@
...
+		_, err := c.GetIndex(w)
+		if err != nil {
+			return nil, err
+		}
+		words = append(words, w)
+	}
+	return words, nil
+}
+
+func (c *WordCodec) WordsToBytes(words []string) ([]byte, error) {
```
rigelrozanski, line 133: Are we enforcing a mandatory number of words for the Recovery phrase? ... oh okay, just read a bit more so will this function validate the recovery phrase if we create a random seed from something else?

rigelrozanski:
Looks good, one overall comment is that typically recovery phrases are assigned and recovered on a one-by-one basis, not as a giant group separated by spaces - I'm not sure what the security implications are though

keys/wordcodec.go
ethanfrey: Yeah, it has a checksum, so we can verify that this phrase came from us (most likely) and is properly decoded.  A given number of words may correspond to two possible byte lengths, so this is super important, not just to detect typos.

ethanfrey: Once this decodes the bytes, it still must pass go-wire, which enforces the size.  ed25519 and secp256k1 have different byte sizes for the private keys.  which means that they have different word lengths. 

I verify the checksum here, I let the go-wire/go-crypto decoding check the proper length.

PR #23, opened 2017-07-22
Shorter recovery seed

ethanfrey:
The recovery seed was 27 words long(!) as I did the most simplistic approach...

Serialize the privkey with go-wire (33 bytes), add 4 byte checksum (37 bytes) and then convert....
37 * 8 / 11 = 26.9 = 27 words

We don't need the whole privkey, but a random seed with 128-bits, which can be used to generate the key, this is all the randomness we can use for the algorithms that have 128-bit security. New approach:

Generate 16 bytes of random data and use that to generate the key.
Append one type-byte (ed25519 or secp256k1) to the end, and sign with crc16 for error checks.
Total length: 19 bytes (19 * 8 / 11 = 13.8 = 14 words)

We don't have the 12 word bip phrases, but we also allow multiplexing and have typo-detection (warm you about a bad word, rather than giving a different key back), at the cost of 2 words. At least it doesn't look some oppressive any more.


keys/cryptostore/holder.go
```diff
@@ -43,12 +43,20 @@ func (s Manager) Create(name, passphrase, algo string) (keys.Info, string, error
 	if err != nil {
 		return keys.Info{}, "", err
 	}
-	key := gen.Generate()
+
+	// 128-bits the the all the randomness we can make use of
```
odeke-em: nit: `the the all`, did you mean 'are all the randomness'?

keys/ecc.go
```diff
@@ -26,6 +28,67 @@ var _ ECC = NoECC{}
 func (_ NoECC) AddECC(input []byte) []byte            { return input }
 func (_ NoECC) CheckECC(input []byte) ([]byte, error) { return input, nil }
 
+// CRC16 does the ieee crc16 polynomial check
+type CRC16 struct {
+	Poly  uint16
+	table *crc16.Table
+}
+
+var _ ECC = &CRC16{}
```
odeke-em: Minor nit:
```go
var _ ECC = (*CRC16)(nil)
```
so that we don't have randomly allocated global objects lingering.

keys/ecc.go
```diff
@@ -26,6 +28,67 @@ var _ ECC = NoECC{}
...
+	// append it to the input
+	output := append(input, check...)
+	return output
+}
+
+func (c *CRC16) CheckECC(input []byte) ([]byte, error) {
+	table := c.getTable()
+
+	if len(input) <= crc16Size {
+		return nil, errors.New("input too short, no checksum present")
```
odeke-em: We could move this out of this function and make a var to avoid allocations on a potential hot path ie every invocation of this function ie
```go
var errInputTooShortNoChecksum = errors.New("input too short, no checksum present")
```
and then in that block:
```go
if len(input) <= crc16Size {
   return nil, errInputTooShortNoChecksum
}
```

keys/ecc.go
```diff
@@ -26,6 +28,67 @@ var _ ECC = NoECC{}
...
+	}
+	cut := len(input) - crc16Size
+	data, check := input[:cut], input[cut:]
+	crc := binary.BigEndian.Uint16(check)
+	calc := crc16.Checksum(data, table)
+	if crc != calc {
+		return nil, errors.New("Checksum does not match")
```
odeke-em: ditto here about moving this to a var so
`errInvalidChecksum`

keys/ecc.go
```diff
@@ -26,6 +28,67 @@ var _ ECC = NoECC{}
...
+	}
+	return data, nil
+}
+
+func (c *CRC16) getTable() *crc16.Table {
+	if c.table == nil {
```
odeke-em: Minor nit: we can reverse conditional this and make for early returns ie
```go
if c.table != nil {
   return c.table
}

if c.Poly == 0 {
   c.Poly = crc16.IBM
}
c.table = crc16.MakeTable(c.Poly)
return c.table
```

keys/cryptostore/holder.go
```diff
@@ -61,15 +69,19 @@ func (s Manager) Create(name, passphrase, algo string) (keys.Info, string, error
...
 	words := strings.Split(strings.TrimSpace(seedphrase), " ")
-	data, err := s.codec.WordsToBytes(words)
+	secret, err := s.codec.WordsToBytes(words)
 	if err != nil {
 		return keys.Info{}, err
 	}
 
-	key, err := crypto.PrivKeyFromBytes(data)
+	l := len(secret)
+	secret, typ := secret[:l-1], secret[l-1]
```
odeke-em: Let's add a comment above that:
```go
// secret is comprised of the actual secret with the type appended
// ie [secret] = [secret] + [type]
```

keys/cryptostore/holder.go
```diff
@@ -43,12 +43,20 @@ func (s Manager) Create(name, passphrase, algo string) (keys.Info, string, error
...
+	secret := crypto.CRandBytes(16)
+	key := gen.Generate(secret)
 	err = s.es.Put(name, passphrase, key)
 	if err != nil {
 		return keys.Info{}, "", err
 	}
-	seed, err := s.codec.BytesToWords(key.Bytes())
+
+	// TODO: clean up type/kind handling with go-data
+	typ := key.Bytes()[0]
```
odeke-em: Let's add a comment here to state that we'll be using secret to carry along the type

odeke-em (changes requested):
@ethanfrey I've posted the first wave of review for things that stood out. After this, I'll give it another review for functional checks. Thanks!

keys/ecc_test.go
```diff
@@ -14,6 +14,9 @@ func TestECCPasses(t *testing.T) {
 
 	checks := []ECC{
```
odeke-em: This list doesn't change between TestECCFails and TestECCPasses, could we instead lift it out to make it a variable that will only need to be changed once?

keys/ecc.go
```diff
@@ -26,6 +28,67 @@ var _ ECC = NoECC{}
...
 
+// CRC16 does the ieee crc16 polynomial check
+type CRC16 struct {
+	Poly  uint16
+	table *crc16.Table
+}
+
+var _ ECC = &CRC16{}
+
+const crc16Size = 2
```
odeke-em: Let's make this more descriptive i.e
```go
const crc16ByteCount = 2
```

keys/cryptostore/encoder_test.go
```diff
@@ -5,15 +5,18 @@ import (
...
 	"github.com/tendermint/go-crypto/keys/cryptostore"
 )
 
 func TestNoopEncoder(t *testing.T) {
 	assert, require := assert.New(t), require.New(t)
 	noop := cryptostore.Noop
 
-	key := cryptostore.GenEd25519.Generate()
-	key2 := cryptostore.GenSecp256k1.Generate()
+	key := cryptostore.GenEd25519.Generate(cmn.RandBytes(16))
```
odeke-em, line 18: So given the fact that even in the tests we are throwing around too many of these
```go
.Generate(cmd.RandBytes(16))
```
perhaps we should make helper functions such as
```go
// GenEd25519Key is a convenience method to generate
// an Ed25519 key from a secret of 16 randomly generated bytes.
func GenEd25519Key() crypto.PrivKey {
   return cryptostore.GenEd25519.Generate(cmn.RandBytes(16))
}

// GenSecp256k1Key is a convenience method to generate
// a Secp256k1 key from a secret of 16 randomly generated bytes.
func GenSecp256k1Key() crypto.PrivKey {
   return cryptostore.GenSecp256k1.Generate(cmn.RandBytes(16))
}
```

This makes for a simpler API and frees up users and ourselves too,
from the burden of having to import different libraries as well as
squinting too hard for the details so we can then replace uses with
```go
key1 := cryptostore.GenEd25519Key()
key2 := cryptostore.GenSecp256k1Key()
```

PR #25, opened 2017-07-26
README: add godoc instead of tedious MD regeneration

### LINKED ISSUE #24: README: use godoc reference instead of hardcoded index/table-of-contents
Currently the README consists of a hardcoded index, listing every exposed function's signature and containing a hardcoded link to the definition
screen shot 2017-07-26 at 3 56 10 pm

There are some problems with this:
1) The function definitions are hardcoded by line number ie
a) Click on https://github.com/tendermint/go-crypto#CRandBytes
screen shot 2017-07-26 at 4 00 23 pm
b) Now click on the link to try to visit the function definition which is hardcoded to https://github.com/tendermint/go-crypto/blob/master/src/target/random.go?s=698:734#L28. That gives a 404
screen shot 2017-07-26 at 4 01 21 pm
2) A mere line change(addition or subtraction) disrupts the definitions
3) It is tedious to try to keep in sync with those definitions as exhibited in point 1) in which the file random.go no longer exists, but the new definition hasn't been changed in the README.

I propose we just switch to using godoc.org https://godoc.org/github.com/tendermint/go-crypto which is built exactly for this purpose so that we can easily navigate through Go source code with ease, they handle all the indexing for us and we don't have to worry at all about moving and keeping definitions in sync.

odeke-em:
Add a godoc reference to delegate indexing and references
to that service, thus keeping our library and usages upto date,
instead of relying on generated markdown that goes stale
especially for changing API signatures, and becomes tedious
where someone has to remember to always regenerate the README
markdown and then commit it.

Fixes #24


ethanfrey:
Looks good. Two points:

1. You can remove `docs` target from the `Makefile` as part of the cleanup.
2. Please make all PRs off develop.  We collect everything there until we do a tagged release.

If you fix those two, I'll happily merge this.  Thanks.

odeke-em:
Ahh I see, thanks @ethanfrey for the updates, let me make those edits.

PR #30, opened 2017-09-06
Develop

ebuchman:


hd/address.go
```diff
@@ -282,9 +282,20 @@ func CalcSha512(buf []byte) []byte {
 }
 
 func ReverseBytes(buf []byte) []byte {
-	res := []byte{}
-	for i := len(buf) - 1; i >= 0; i-- {
-		res = append(res, buf[i])
+	var res []byte
```
ebuchman, line 285: Maybe this belongs in tmlibs/common/bytes.go?

keys/cryptostore/holder.go
```diff
@@ -43,12 +43,21 @@ func (s Manager) Create(name, passphrase, algo string) (keys.Info, string, error
...
+	key := gen.Generate(secret)
 	err = s.es.Put(name, passphrase, key)
 	if err != nil {
 		return keys.Info{}, "", err
 	}
-	seed, err := s.codec.BytesToWords(key.Bytes())
+
+	// we append the type byte to the serialized secret to help with recovery
+	// ie [secret] = [secret] + [type]
+	typ := key.Bytes()[0]
```
ebuchman, line 57: seems kind of hacky. is there not a better way to get the type byte?

ebuchman (commented):
Some nice cleanup

keys/cryptostore/holder.go
ethanfrey: there isn't, i was thinking to add this to go-wire, but it is not public afaik

hd/address.go
ethanfrey: Yeah, a nice reverse function... sounds good

Issue #31, opened 2017-09-07
helpers: provide helpers to generate keys


odeke-em:
Am currently writing code to test out the bug fix for https://github.com/tendermint/tendermint/issues/514 in which we have a dishonest peer requesting a non-existent block height from a reactor and that requires a whole lot of setup.

I currently see that we have helpers for PrivKeyEd25519
https://github.com/tendermint/go-crypto/blob/1bc8de4caa844f8b64c120e65b047898f22b7f3e/priv_key.go#L99-L114

However, trying to generate the public key using our package is quite a bit of work, IMHO we should make it simple to generate a public Ed25519 key that conforms to our package and not just to the Go crypto library. I suspect some other developers might need this in the future.

ethanfrey:
What about `GenPrivKeyEd25519().PubKey()`?

That's what I do, and your logic would do under the hood.  Not too much work, no?

odeke-em:
Thanks @ethanfrey, and my apologies for the late reply, I was off from here since the weekend, but am back.

As per your recommendation and exhibit above, I've opened #35 in which we can provide a high level view of docs for go-crypto to easily transfer such knowledge to other developers and provide some more documentation. I'll close this issue in favor of the mentioned one.

PR #32, opened 2017-09-07
GenEd25519KeyPair* helpers added

odeke-em:
Fixes https://github.com/tendermint/go-crypto/issues/31

Add a helper to be able to generate Ed25519 keys directly
from our crypto package, and with these helpers we give
more context to a previously undocumented and hanging
code statement in which while generating a private key
we invoked ed25519.MakePublicKey, which is correct but
seemed hanging. With this change, the helpers are:
```go
func GenEd25519KeyPair() (priv PrivKeyEd25519, pub PubKeyEd25519)
func GenEd25519KeyPairFromSecret(secret []byte) (priv PrivKeyEd25519, pub PubKeyEd25519)
```

which are invoked by:
- GenPrivKeyEd25519()
- GenPrivKeyEd25519FromSecret(secret []byte)
respectively and now with the pairings give more comfort knowing that
each private key for completion must be generated with its public key,
but we'll chose to discard the public key.


ethanfrey:
Huh?  I was used to generating a private key, and later always just called privKey.PubKey() when I needed it.

The code looks correct, still not sure it makes life easier. Can you reference another package that uses this library that would be improved by this change?

ebuchman:
Hmm I agree with Frey here. Why do this when we can just `privKey.PubKey()` ?

odeke-em:
Thanks for the input @ethanfrey @ebuchman, sorry for the late reply, been off for a bit.
So unfortunately that wasn't documented(at least I did not find it easily).

odeke-em:
@ethanfrey in regards to packages that would need this, I needed this for this PR https://github.com/tendermint/tendermint/pull/648/files#diff-4b45e760192d8a11e9b4e278a8c8ea32R115

ebuchman:
Is the following not sufficient:

```
pubKey := crypto.GenPrivKeyEd25519().PubKey()
```
?

odeke-em:
@ebuchman it is sufficient indeed, but now that y'all pointed me to it. That didn't seem documented, at least I couldn't easily find a reference to it.

ebuchman:
So let's close this an open an issue to create a doc.go explaining how this package works?

odeke-em:
SGTM, great idea!

odeke-em:
Thanks @ebuchman and @ethanfrey for the review and the direction, I've opened https://github.com/tendermint/go-crypto/issues/35 which supersedes this PR.

PR #37, opened 2017-09-13
Nano Support

ethanfrey:
All drivers working with alpha cosmos app. Can sign keys.  Want review and feedback.

This code was integrated with the CLI in cosmos-sdk and can successfully sign basecoin transactions. There is still some work to do for proper key management on the nano (not just one static key per device session), but for proof of concept, the code works.

One big issue I have is the PubKey/PrivKey interface. Naturally, we assumed the methods like `priv.PubKey()` and `priv.Sign(..)` would never fail, as it was math.  Now there is I/O involved, and we get errors if the ledger is disconnected or the cosmos app is not open. Unfortunately, there is no way to return errors in the current interface, so I have to panic on those errors :( Changing the interface would break lots of code, and is not undertaken lightly.

Can I get some feedback here on how to do this proper?  The success case works fine, the error handling just sucks.


odeke-em:
Hey @ethanfrey, short of fixing up all the signatures to return `(crypto.PrivKey, error)`,
here is a hack that I think could help:
a) instead of making `genLegder` a GenFunc, make it a struct that can save state i.e the last error
e.g
```go
type genLedger struct {
   mu sync.Mutex
   lastErr error
}
```
b) On encountering any error, instead of panicking, save that `lastErr` e.g `gl.lastErr = err` then return normally
c) Make `genLedger` implement the LastErrorer `LastError() error` interface which evicts the last error on every read. This ensures that we capture the last encountered error only once, sort of as if it were a function call. Thus
```go
func (gl *genLedger) LastError() error {
   gl.mu.Lock()
   var err error = gl.lastErr
   gl.lastErr = nil
   gl.mu.Unlock()
   return err
}

func (gl *genLedger) setLastErr(err error) {
   gl.mu.Lock()
   gl.lastErr = err
   gl.mu.Unlock()
}

func (gl *genLedger) Generate(secret []byte) crypto.PrivKey {
   key, err := nano.NewPrivKeyLedger()
   if err != nil {
       gl.setLastErr(err)
   }
   return key
}

type LastErrorer interface {
   LastError() error
}
```

and then for usage:
```go
key := gen.Generate()
if le, ok := gen.(LastErrorer); ok {
   if err := le.LastError(); err != nil {
     // Handle the error here
   }
}
```
or you can make that a helper inside the code that needs it
```go
func generateKey(gen Generator) (crypto.PrivKey, error) {
   key := gen.Generate()
   le, ok := gen.(LastErrorer)
   if !ok {
      return key, nil
   }
   err := le.LastError()
   return key, err
}
```

and then use it normally in any of the usages as
```go
key, err := generateKey(gen)
```
which will prevent mortally breaking this repo. With the mentioned pattern, we can slowly refactor all our code and eventually make all usages return `(crypto.PrivKey, error)` eventually achieving the proper signature.

Please let me know what you think.

ebuchman:
What if instead of exposing this thing as a crypto.PrivKey we abstract the Signer, and give it a `Sign( ...) (crypto.Signature, error)`. Would this solve the problem ? Then the Signer could have additional rules that might cause it to err too. The consumers can just check for errors on calling Sign. Everything else should work fine, and constructors should populate pubkeys and return errors if they can't.

For instance this is how the PrivValidator works in Tendermint, via a Signer interface with a Sign function. And here's the commit where I just added the error: https://github.com/tendermint/tendermint/commit/90c0267bc1a490d0501c7fa1c5916132f4d6be1e

ethanfrey:
Thank you both for the ideas. I will take a deeper look at the code you both suggested on Monday. But happy to see some solutions.

keys/cryptostore/generator.go
```diff
@@ -32,12 +36,25 @@ func genSecp256(secret []byte) crypto.PrivKey {
...
+	return key
+}
+
 func getGenerator(algo string) (Generator, error) {
 	switch algo {
 	case crypto.NameEd25519:
 		return GenEd25519, nil
 	case crypto.NameSecp256k1:
 		return GenSecp256k1, nil
+	case nano.NameLedger:
```
zmanian: IMO this should NameLedgerEd25519

The ledger can support more key types in the future.

ethanfrey: Sounds good, done.

ethanfrey:
I fixed up the Generator signature to allow for errors when creating new keys in the key manager. I also extended PrivKeyFromBytes to allow a PrivKey to validate its correctness upon loading, as an optional check. Only done by nano by pinging the app, but ed25519 could make sure the priv/pub key parts match for example, as we serialize them both.

Now errors are returned, not panics, in the key manager upon creating a key on the ledger or signing with the ledger, if it is not plugged in with the cosmos app running.

nano/sign.go
```diff
@@ -0,0 +1,83 @@
...
+	App       = 0x80
+	Init      = 0x00
+	Update    = 0x01
+	Digest    = 0x02
+	MaxChunk  = 253
+	KeyLength = 32
+	SigLength = 64
+)
+
+var separator = []byte{0, 0xCA, 0xFE, 0}
```
wrl, line 20: fine for now, but we should eventually remove this separator in favor of just having hardcoded slice offsets.

wrl: (just a heads-up)

keys/cryptostore/generator.go
```diff
@@ -4,53 +4,86 @@ import (
...
+	"github.com/tendermint/go-crypto/nano"
 )
 
 var (
 	// GenEd25519 produces Ed25519 private keys
 	GenEd25519 Generator = GenFunc(genEd25519)
 	// GenSecp256k1 produces Secp256k1 private keys
 	GenSecp256k1 Generator = GenFunc(genSecp256)
+	// GenLedger used Ed25519 keys stored on nano ledger s with cosmos app
+	GenLedger Generator = GenFunc(genLedger)
```
ebuchman: Do we want to support Ledger-Secp256k1? Eg. the fundraiser keys

nano/keys.go
```diff
@@ -0,0 +1,283 @@
...
+	if !pub.Equals(pk.CachedPubKey) {
+		return errors.New("ledger doesn't match cached key")
+	}
+	return nil
+}
+
+// AssertIsPrivKeyInner fulfils PrivKey Interface
+func (pk *PrivKeyLedger) AssertIsPrivKeyInner() {}
+
+// Bytes fulfils pk Interface - no data, just type info
```
ebuchman: what's this mean just type info?

nano/keys.go
```diff
@@ -0,0 +1,283 @@
...
+func (pk *PrivKeyLedger) Bytes() []byte {
+	return wire.BinaryBytes(pk.Wrap())
+}
+
+// Sign calls the ledger and stores the pk for future use
+func (pk *PrivKeyLedger) Sign(msg []byte) crypto.Signature {
```
ebuchman: Let's add method comment here `XXX/TODO: it panics if there's an error`

nano/keys.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+	key, _, err = signLedger(dev, []byte{0})
+	if err != nil {
+		return key, errors.New("Please open cosmos app on the ledger")
+	}
+	return key, err
+}
+
+// Equals fulfils PrivKey Interface
+// TODO: needs to be fixed
```
ebuchman: should probably be addressed!

nano/keys.go
```diff
@@ -0,0 +1,283 @@
...
+// VerifyBytes uses the normal Ed25519 algorithm but a sha512 hash beforehand
+func (pk PubKeyLedger) VerifyBytes(msg []byte, sig crypto.Signature) bool {
+	hmsg := hashMsg(msg)
+	return pk.PubKeyEd25519.VerifyBytes(hmsg, sig)
+}
+
+// Equals implements PubKey interface
+func (pk PubKeyLedger) Equals(other crypto.PubKey) bool {
+	if ledger, ok := other.Unwrap().(PubKeyLedger); ok {
+		return bytes.Equal(pk.PubKeyEd25519[:], ledger.PubKeyEd25519[:])
```
ebuchman: could we not do `pk.PubKeyEd25519.Equals(ledger.PubKeyEd25519)` ?

nano/keys.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+	return device, err
+}
+
+func signLedger(device *ledger.Ledger, msg []byte) (pk crypto.PubKey, sig crypto.Signature, err error) {
+	var resp []byte
+
+	packets := generateSignRequests(msg)
+	for _, pack := range packets {
+		resp, err = device.Exchange(pack, 100)
```
ebuchman: magic number

ebuchman (approved):
Seems relatively straight forward. Haven't consulted the manual or tested it yet but otherwise looks good.

The ethanfrey/ledger library could use some comments :)

keys/cryptostore/generator.go
ethanfrey: Cosmos app doesn't support secp256k1, but i updated all the names to clarify this is ed25519

nano/keys.go
ethanfrey: yes, you are right. Except we need to add a Wrap() to make it compile. Replaced this logic.

nano/keys.go
ethanfrey: This was before I stored the cached key, as it was regenerated every command. Now fixed it.

nano/keys.go
ethanfrey: Added

nano/keys.go
ethanfrey: Fixed comment

keys/cryptostore/generator.go
ebuchman: Ok - the fundraiser keys were secp256k1. So we need the app to support it.

wrl: @ebuchman we touched on this on slack briefly – do we have any need for the ledger to support secp256k1, or will it remain ed25519-only?

ebuchman: it would be good too.

PR #40, opened 2017-09-15
doc.go file started

### LINKED ISSUE #35: docs: implement a doc.go file to provide high level usage and code usage of this package
This issue supersedes https://github.com/tendermint/go-crypto/pull/32 and its issue https://github.com/tendermint/go-crypto/issues/31 in which it I had to mistakenly dig through the source code of `GenEd25519Key` and extracted out a few helpers to retrieve the public key.

However, @ethanfrey and @ebuchman gave me guidance on the PR #32 that we can in fact invoke 
`GenEd25519Key().PubKey()` which will retrieve the public key. 

A doc.go file is a nice way for us to cut the bus factor for our code as well as make it really simple for other developers to pick out and reference code. It also gives a nicer higher level introduction to our packages and provides quick documentation that's viewable by godoc. We talked about this offline in the Slack channel as per exhibit[0]

exhibit[0]
If so, we can do such a thing as https://github.com/golang/go/blob/0202aa8ec79bf700732812624eedcec3e6e300ad/src/net/http/doc.go#L5-L107 which then gives the `Overview` at https://golang.org/pkg/net/http/
which gives

https://godoc.org/github.com/golang/go/src/net/http

odeke-em:
Fixes #35
Updates https://github.com/tendermint/coding/issues/27

Started a doc.go file to provide an overview/high level
dive into the functionality of this repo.

Also added an example_test.go file in which we can
put end-to-end code examples/actual usage patterns
that can be copied and pasted and will always have to
compile when tests are run to ensure that we don't regress.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -1,0 +1,48 @@
+/*
+go-crypto is a customized/convenience cryptography package
+for supporting Tendermint.
+
+It wraps select functionality of equivalent functions in the
+Go standard library, for easy usage with our libraries.
+
+Keys:
+
+All key generation functions return an instance of the PrivKey interface
+which implements methods
+
+    AssertIsPrivKeyInner()
+    Bytes() []byte
+    Sign(msg []byte) Signature
+    PubKey() PubKey
+    Equals(PrivKey) bool
+    Wrap() PrivKey
+
+From the above method we can:
+a) Retrieve the public key if needed
+
+    pubKey := key.PubKey()
+
+For example:
+    privKey, err := crypto.GenPrivKeyEd25519()
+    if err != nil {
+	...
+    }
+    pubKey := privKey.PubKey()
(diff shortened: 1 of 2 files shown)


odeke-em:
Exhibit of the godoc output
screen shot 2017-09-15 at 5 09 53 pm

odeke-em:
/cc @ebuchman @zramsay

With the initial file added, we can then add more information there piecemeal.

adrianbrink (approved):
Seems like a good idea to have doc.go files at the top level.

zramsay (approved):

odeke-em:
Thanks for the reviews @adrianbrink and @zramsay, I'll merge it then.

ebuchman:
Thanks odeke!

PR #42, opened 2017-10-24
Bcrypt usability

ethanfrey:
I extracted the cost as a parameter. The current value (14) is nice and secure, but kills the UX. Signing a tx (or returning a bad password error) takes around 1 second just due to the bcrypt decoding of the key on my desktop (i7), slower on other computers. Tests are super slow.

Reducing this to 12 (a factor of 4) makes it much more usable, and I think the security is still quite high. Remember that we decrypt the key every time we use it to avoid keeping it in memory and passing it outside the key manager.

If you think I am destroying the security of our key system, please reject this.

### DIFF
--- a/keys/cryptostore/encoder.go
+++ b/keys/cryptostore/encoder.go
@@ -7,6 +7,20 @@ import (
 	"github.com/tendermint/go-crypto/bcrypt"
 )
 
+const (
+	// BcryptCost is as parameter to increase the resistance of the
+	// encoded keys to brute force password guessing
+	//
+	// Jae: 14 is good today (2016)
+	//
+	// Ethan: loading the key (at each signing) takes a second on my desktop,
+	// this is hard for laptops and deadly for mobile. You can raise it again,
+	// but for now, I will make this usable
+	//
+	// TODO: review value
+	BCryptCost = 12
+)
+
 var (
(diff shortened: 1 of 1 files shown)


melekes (approved):

adrianbrink (approved):

mveytsman:
Just out of curiosity, what's the threat model with not storing the key in memory inside the key manager while the program is running?

Are we protecting against malware that has memory read access but not at the point that a transaction is happening? or cold boot attacks?

ebuchman:
> Are we protecting against malware that has memory read access but not at the point that a transaction is happening? or cold boot attacks?

Good question. Seems standard for the key to be kept in memory for a short time, for convenience sake. The tradeoff is there's a longer period during which someone could hit you with a cold boot attack, vs. having to time it for right after you sign.

Malware that doesn't have read access at the point of the tx happening seems unlikely.

Maybe we should consider storing the key in memory unlocked for a few minutes at a time?

mveytsman:
From a user's perspective, I think it's okay to wait a second to unlock the wallet and then have the key in memory for a configurable amount of time (i.e 1 minute, 10 minutes, 1 hour, until program is quit)

Also, we're relying on Go's garbage collection to clear the keys out of memory. It's worth researching if there's a deterministic way to make sure an address is zeroed out.

Issue #43, opened 2017-10-26
ensure that private key comparisons are constant time comparisons to avoid timing attacks


odeke-em:
Let's inspect our code and ensure that any private key or HMAC comparisons are always constant time comparisons instead of using `bytes.Equal` which is susceptible to timing/side channel attacks. When cryptographers audit our code, most likely the keen ones will point out the same problem out.

PR #44, opened 2017-10-26
make PrivateKey + Signature comparisons use constant time comparisons

odeke-em:
Fixes https://github.com/tendermint/go-crypto/issues/43

Avoid susceptibility to timing/side channel attacks by ensuring
that private key and signature comparisons use
`subtle.ConstantTimeCompare`
instead of
`bytes.Equal`

### DIFF
--- a/priv_key.go
+++ b/priv_key.go
@@ -1,7 +1,7 @@
 package crypto
 
 import (
-	"bytes"
+	"crypto/subtle"
 
 	secp256k1 "github.com/btcsuite/btcd/btcec"
 	"github.com/tendermint/ed25519"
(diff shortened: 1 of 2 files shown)


ebuchman:
Thanks @odeke-em . Seems like a good idea.

But does it matter for signatures? And where do we even use privKey.Equals ?

odeke-em:
@ebuchman, yeah for signatures I infer that from the case that those are like authenticated messages, but I perhaps am wrong that it doesn't matter.

In regards to where we use PrivKey.Equals, interestingly I don't seem to see usages in our repos.
If the usage/correctness doesn't matter, then perhaps we should delete those otherwise it is something that will get the ears of any auditor burning. I found this by isolating package by package and reading through to ensure just sort of correctness of the code.

ebuchman:
Hm, seems maybe we can eliminate both Equals ? I see no reason to ever be checking privkey equality like that, and same for signatures? Are we using that anywhere?

zramsay:
status?

ebuchman:
Replaced by https://github.com/tendermint/go-crypto/pull/53/files

ebuchman:
@odeke-em note `subtle` uses `1` to denote equality: https://golang.org/pkg/crypto/subtle/#ConstantTimeCompare
 
Also, is there some reason the signatures need to be checked in constant time?

odeke-em:
Thank you @ebuchman for the refurbish! Mother of God, why did I use `0` instead? Moreover I was somewhat involved in reviewing the original Go package signature IIRC, how could I not catch that.

ebuchman:
oh cool - agl doesn't think verify should be in constant time either: https://github.com/golang/crypto/commit/c412588e259ef2d488e15762500456cd3b171bad#diff-8183aa2c95dc1385c8f58bf1917fcb0b

odeke-em:
Gotcha gotcha, thanks for the discussion and you called this about signatures not needing constant time comparisons.

PR #46, opened 2017-10-26
Develop

ebuchman:


keys/cryptostore/encoder.go
```diff
@@ -16,45 +32,55 @@ var (
...
+	Decrypt(saltBytes []byte, encBytes []byte, passphrase string) (privKey crypto.PrivKey, err error)
 }
 
 type secretbox struct{}
 
-func (e secretbox) Encrypt(key crypto.PrivKey, pass string) ([]byte, error) {
-	if pass == "" {
-		return key.Bytes(), nil
+func (e secretbox) Encrypt(privKey crypto.PrivKey, passphrase string) (saltBytes []byte, encBytes []byte, err error) {
+	if passphrase == "" {
```
adrianbrink, line 42: Can't this be deleted, since we don't want users to encrypt keys without a passphrase?

ethanfrey: We break enough stuff with our encryption library. This is the only way to save keys between versions... export unencrypted and import unencrypted.
We should expose them and use them to migrate keys. Until production release and a stable go-crypto

ebuchman: Does this mean its currently possible to migrate keys? We should defntly be trying to support that asap if possible ...

PR #50, opened 2017-10-29
Bycrypt upgrade

adrianbrink:
This is a backwards compatible upgrade to use salts. Old existing keys that were stored without a salt are still decryptable. However, any new key generated and stored will use a salt. Also, it is not possible to create new keys without a passphrase. Old keys without a passphrase are still decryptable.

@ethanfrey Please check that you are okay with the change to passphrases. I disabled two tests that try and create new keys without a passphrase.

@ebuchman As far as I can tell this shouldn't break any old key. Clients will still have to update their code, but when they want to import a key that was exported without a salt they can pass `nil` for the salt. So even a user that exported his keys before the change, should be able to import the same key.


melekes (approved):

zramsay:
merge?

ebuchman:
handled in https://github.com/tendermint/go-crypto/pull/58/

PR #51, opened 2017-11-01
Keybase

jaekwon:
Refactoring go-crypto/keys to be much simpler, and also to use go-db.

Also see https://github.com/tendermint/tmlibs/pull/72

Please use the logic from https://github.com/tendermint/mintkey/blob/master/cmd/mintkey/common.go to encrypt and armor private keys.  I really don't think we need to create DB/store wrapper abstractions to make this work.

Some other feedback:
* We should avoid calling things "Manager"
* We should not use single letter receiver names, like func (s Something) MyFunc()... it makes refactoring s. -> kb. more difficult than it should be.  https://github.com/tendermint/coding/commit/0057139c237c888a0a31ca75714bc6d7489b8686#diff-87e87465831927a211c5a12fa8ace289


jaekwon:
Ah, and we don't need all those GenEd25519.Generate type functions... it could be as simple as a switch statement in the dbKeybase impl.

ethanfrey:
Jae, I know you want to simplify and I think using the level db backend is nice.

However, a few main points:

1. You destroyed all backwards compatibility with the key files, which I get shit for on every gaia release. Talk to Bucky on that one, I don't promise any backwards compatibilty.

2. You removed all the generators, while they are still relied upon for creating keys.  If you think it can be done simpler, please do that.

3. Changing the signing interface breaks `go-crypto/client/commands/keys`. When refactoring libs, please only break one at a time. Like make changes in go-crypto and see that you can get a branch of tendermint:develop and cosmos-sdk:develop to compile again. Changing this API implies a major refactor of all signing in the cosmos-sdk.

Also, please make sure this works (with proper error handling) with the nano leger, I have a ledger with the app installed here, so we can test it. I see lots of deleted test code, which always makes me very nervous.

adrianbrink:
@jaekwon @ebuchman We should make a decision on backwards compatibility. 

I am personally in favour of having no backwards compatibility for anything from APIs to keyfiles until after launch. For now we can expect most people to upgrade and maintaining backwards compatibility sometimes introduces these weird code paths that we can otherwise avoid.

ethanfrey:
It all compiles and almost all tests pass. There is one issue I cannot resolve, as I cannot figure out how to properly use the dbm.Iterator.

If I call Next(), it will return false if I was at the end, or true and advance one step.
This is fine if there is at least one element, which I read, then call next until it returns false.

However, if there is no elements in the response, reading the first element will panic. If I call next to check if there was an element before reading it, it is safe, but I will always miss the first element.

@jaekwon I know you have a PR on tmlibs that makes this nicer. Maybe I wait until then, or you can fix this code to make the tests work.

https://github.com/tendermint/go-crypto/blob/keystore/keys/keybase.go#L86-L102

ethanfrey:
I am sure it also all needs a review and polish, but at least it is quite close to working code now.

ethanfrey:
vetshadow and zach's linting obsession kill all tests when the variable name is the same as a package (as it shadows the previous value).

I don't like to change it, but needed to change one name, so changed the name we saw less.

rigelrozanski:
oh i c

PR #59, opened 2017-12-30
Sdk2 cleanup

ebuchman:


keys/keybase.go
```diff
@@ -176,65 +178,62 @@ func (kb dbKeybase) Delete(name, passphrase string) error {
...
-	kb.db.DeleteSync(pubName(name))
-	kb.db.DeleteSync(privName(name))
-	kb.writeKey(key, name, newpass)
+	// Generate the public bytes and the encrypted privkey
+	public := info(name, key)
+	private := encryptArmorPrivKey(key, newpass)
+
+	// We must delete first, as Putting over an existing name returns an error.
+	// Must be done atomically with the write or we could lose the key.
+	batch := kb.db.NewBatch()
```
ebuchman, line 198: fsdb doesn't support batch. need to do something different.

PR #61, opened 2018-01-02
PrivKey*: fix UnmarshalJSON to return err ASAP avoiding unconditional copy call

odeke-em:
Noticed while reading through
https://github.com/tendermint/tendermint/pull/1048/files#diff-072f7b0dc4657b530eec2f4a5ca94e5bR82

that we weren't checking for zeroness, but then the UnmarshalJSON
methods of:
* PrivKeyEd25519
* PrivKeySecp256k1
would return the error but unconditionally invoked copy even
on zeroed bytes, hence return early if we encounter an error
and only invoke copy if we've passed.

### DIFF
--- a/priv_key.go
+++ b/priv_key.go
@@ -85,9 +85,11 @@ func (p PrivKeyEd25519) MarshalJSON() ([]byte, error) {
 
 func (p *PrivKeyEd25519) UnmarshalJSON(enc []byte) error {
 	var ref []byte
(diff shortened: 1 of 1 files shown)


zramsay:
fee free to disable `vetshadow`

ebuchman:
Nice find, thanks! But no longer relevant post amino :)

PR #63, opened 2018-01-14
Use new go-wire; PubKey etc are interfaces; Keybase refactor

jaekwon:
Nano ledger shouldn't be a PubKey...  PubKey is something serializable.

Keybase could be extended to have more methods that deal with attaching devices with a name.  Sign() could return either a nil PubKey, or figure out the PubKey from the device, or we could add more methods.  In any case, HSMs are orthogonal to PubKey/PrivKey/Signature.


jaekwon:
Merging to make consistent with sdk2 everywhere.  Lets continue to discuss here.

PR #64, opened 2018-01-14
Merge sdk2 into develop

ebuchman:


keys/keybase.go
```diff
@@ -32,58 +26,59 @@ func New(db dbm.DB, codec words.Codec) dbKeybase {
...
+		return Info{}, "", err
 	}
 
 	// encrypt and persist the key
-	public := kb.writeKey(key, name, passphrase)
+	info := kb.writeKey(priv, name, passphrase)
+
+	// we append the type byte to the serialized secret to help with
+	// recovery
+	// ie [secret] = [type] + [secret]
```
ebuchman, line 49: why do this? it seems wrong to influence the words in any way. Why not just pass in the algo explicitly ? This seems like unecessary unstandard weirdness ...

keys/keybase.go
```diff
@@ -32,58 +26,59 @@ func New(db dbm.DB, codec words.Codec) dbKeybase {
...
-	public := kb.writeKey(key, name, passphrase)
-	return public, nil
+	seed := strings.Join(words, " ")
+	return info, seed, err
 }
 
-// SeedToPrivKey returns the private key corresponding to a seedphrase
-// without persisting the private key.
-// TODO: enable the keybase to just hold these in memory so we can sign without persisting (?)
-func (kb dbKeybase) SeedToPrivKey(algo, seedphrase string) (crypto.PrivKey, error) {
```
ebuchman, line 75: What if we just want to use a private key from seed without persisting it to disk? that's what this function is for ...

keys/keybase.go
```diff
@@ -92,156 +87,116 @@ func (kb dbKeybase) List() ([]Info, error) {
...
+	bz := kb.db.Get(infoKey(name))
+	if bz == nil {
+		return "", errors.New("No key to export with name " + name)
 	}
-	armorBytes := encryptArmorPrivKey(key, transferpass)
-	return []byte(armorBytes), nil
+	return armorInfoBytes(bz), nil
 }
 
-// Import accepts bytes generated by Export along with the same transferpass.
```
ebuchman, line 149: Why remove comments?

keys/keybase.go
jaekwon: it's a UX thing.  I don't want to have to remember what the words were for.

jaekwon: It'll just confused the user.  Maybe we can just prepend a single word that represents the type of key, so it's 13 words?

keys/keybase.go
jaekwon: who calls it?  I couldn't find any usage needs.

keys/keybase.go
jaekwon: The old comment was wrong, and the API I think is self explanatory for now...

keys/keybase.go
ebuchman: We haven't exposed it yet but we should. Also the UI is now asking for it (https://github.com/cosmos/cosmos-sdk/issues/301) . We defintely should have a way to enter a seed and use the key without it being persisted to disk.

keys/keybase.go
ebuchman: Should we not by convention have a proper godoc comment on every exposed item ? Even if it's obvious?

keys/keybase.go
ebuchman: lets do eg. `ed25519{ house table chair ... }`

Issue #65, opened 2018-02-10
Integration with bech32cosmos


zmanian:
Feature:

Go-crypto should be integrated with [bech32cosmos](https://github.com/cosmos/bech32cosmos)

String representation of cryptographic objects that come from go-crypto or are deserialized by go-crypto should only be provided in bech32cosmos form.

I think this is the right way to approach https://github.com/cosmos/cosmos-sdk/issues/263 but looking for confirmation.

jaekwon:
Too many things... lets keep the discussion on https://github.com/cosmos/cosmos-sdk/issues/263 ,
and there is a PR at https://github.com/tendermint/go-crypto/pull/66

PR #66, opened 2018-02-12
Add support for bech32cosmos

zmanian:
Implements #65 

Adds bech32cosmos as dependency in glide
Turn Address into a struct with an anonymous field to allow for method overrides
Added change String function to do bech32

### DIFF
--- a/pub_key.go
+++ b/pub_key.go
@@ -31,7 +31,10 @@ func (a *Address) UnmarshalJSON(data []byte) error {
 		return fmt.Errorf("%s is not CSMSADDR the Cosmos Address identifier", readable)
 	}
 
-	cdc.UnmarshalBinary(deserialized, a)
+	err = cdc.UnmarshalBinary(deserialized, a)
+	if err != nil {
+		return err
+	}
 	return nil
 }
 
@@ -41,7 +44,9 @@ func (a *Address) MarshalJSON() ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	bech, err := bech32cosmos.Encode("CSMSADDR", marshaled)
+	conv, err := bech32cosmos.ConvertBits(marshaled, 8, 5, true)
+
+	bech, err := bech32cosmos.Encode("CSMSADDR", conv)
 	if err != nil {
 		return nil, err
 	}
@@ -51,13 +56,38 @@ func (a *Address) MarshalJSON() ([]byte, error) {
 func (a *Address) String() string {
 	marshaled, err := cdc.MarshalBinary(a)
 	if err != nil {
-		return err.Error()
+		return "go-wire err:" + err.Error()
 	}
-	bech, err := bech32cosmos.Encode("CSMSADDR", marshaled)
+	conv, err := bech32cosmos.ConvertBits(marshaled, 8, 5, true)
+
+	bech, err := bech32cosmos.Encode("CSMSADDR", conv)
 	if err != nil {
-		return err.Error()
+		return "bech32cosmos err:" + err.Error()
 	}
 	return bech
+}
+
+func (a *Address) FromString(str string) error {
+	readable, deserialized, err := bech32cosmos.Decode(str)
+	if err != nil {
+		return err
+	}
+	if strings.ToUpper(readable) != "CSMSADDR" {
+		return fmt.Errorf("%s is not CSMSADDR the Cosmos Address identifier", readable)
+	}
+
+	deserialized, err = bech32cosmos.ConvertBits(deserialized, 5, 8, true)
+
+	err = cdc.UnmarshalBinary(deserialized, a)
+
+	if err != nil {
+		return err
+	}
+
+	if err != nil {
+		return err
+	}
+	return nil
 }
 
 func PubKeyFromBytes(pubKeyBytes []byte) (pubKey PubKey, err error) {
@@ -123,7 +153,9 @@ func (pubKey PubKeyEd25519) String() string {
 	if err != nil {
 		return err.Error()
 	}
-	bech, err := bech32cosmos.Encode("CSMSPUB", marshaled)
+	conv, err := bech32cosmos.ConvertBits(marshaled, 8, 5, true)
+
+	bech, err := bech32cosmos.Encode("CSMSPUB", conv)
 	if err != nil {
 		return err.Error()
 	}
@@ -189,7 +221,9 @@ func (pubKey PubKeySecp256k1) String() string {
 	if err != nil {
 		return err.Error()
 	}
-	bech, err := bech32cosmos.Encode("CSMSPUB", marshaled)
+	conv, err := bech32cosmos.ConvertBits(marshaled, 8, 5, true)
+
+	bech, err := bech32cosmos.Encode("CSMSPUB", conv)
 	if err != nil {
 		return err.Error()
 	}
--- a/pub_key.go
+++ b/pub_key.go
@@ -20,6 +20,7 @@ import (
 // Use an alias so Unmarshal methods (with ptr receivers) are available too.
 type Address struct {
 	cmn.HexBytes
+	humanReadable string
 }
 
 func (a *Address) UnmarshalJSON(data []byte) error {
@@ -27,9 +28,7 @@ func (a *Address) UnmarshalJSON(data []byte) error {
 	if err != nil {
 		return err
 	}
-	if strings.ToUpper(readable) != "CSMSADDR" {
-		return fmt.Errorf("%s is not CSMSADDR the Cosmos Address identifier", readable)
-	}
+	a.humanReadable = readable
 
 	err = cdc.UnmarshalBinary(deserialized, a)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


pub_key.go
```diff
@@ -4,6 +4,9 @@ import (
 	"bytes"
 	"crypto/sha256"
 	"fmt"
+	"strings"
+
+	"github.com/cosmos/bech32cosmos/go"
```
adrianbrink: Collapse this into the other external imports

pub_key.go
```diff
@@ -15,7 +18,77 @@ import (
...
+type Address struct {
+	cmn.HexBytes
+}
+
+func (a *Address) UnmarshalJSON(data []byte) error {
+	readable, deserialized, err := bech32cosmos.Decode(string(data))
+	if err != nil {
+		return err
+	}
+	if strings.ToUpper(readable) != "CSMSADDR" {
```
adrianbrink: Mh, I'm wondering whether we want to enforce the address in go-crypto. I can imagine that other zones would like to use a very similar bech encoding scheme. We should allow users to define their own `readable`.

pub_key.go
```diff
@@ -15,7 +18,77 @@ import (
...
+
+	err = cdc.UnmarshalBinary(deserialized, a)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (a *Address) MarshalJSON() ([]byte, error) {
+
```
adrianbrink: No line break

pub_key.go
```diff
@@ -15,7 +18,77 @@ import (
 // An address is a []byte, but hex-encoded even in JSON.
 // []byte leaves us the option to change the address length.
 // Use an alias so Unmarshal methods (with ptr receivers) are available too.
-type Address = cmn.HexBytes
+type Address struct {
+	cmn.HexBytes
```
adrianbrink, line 21: We should track the human readable part here.

pub_key.go
```diff
@@ -134,9 +240,41 @@ func (pubKey PubKeySecp256k1) VerifyBytes(msg []byte, sig_ Signature) bool {
...
 
 func (pubKey PubKeySecp256k1) String() string {
-	return fmt.Sprintf("PubKeySecp256k1{%X}", pubKey[:])
+	marshaled, err := cdc.MarshalBinary(pubKey)
+	if err != nil {
+		return err.Error()
+	}
+	conv, err := bech32cosmos.ConvertBits(marshaled, 8, 5, true)
+
+	bech, err := bech32cosmos.Encode("CSMSPUB", conv)
```
adrianbrink: Same here. I don't think we should enforce the human-readable part.

pub_key.go
zmanian: So where does the readable value get injected from?

ebuchman: Think it makes sense to unmarshal the readable and the data into the `Address` struct. Then consumers of go-crypto can enforce what they need to about the addresses, and we can do things like NewAccountAddress and NewValidatorAddress

pub_key.go
```diff
@@ -15,7 +18,77 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+	return []byte(bech), nil
+}
+
+func (a *Address) String() string {
+	marshaled, err := cdc.MarshalBinary(a)
+	if err != nil {
+		return "go-wire err:" + err.Error()
```
ebuchman: panic ?

pub_key.go
```diff
@@ -15,7 +18,77 @@ import (
...
+	}
+
+	err = cdc.UnmarshalBinary(deserialized, a)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (a *Address) MarshalJSON() ([]byte, error) {
```
ebuchman, line 39: can't this just call String()?

pub_key.go
```diff
@@ -15,7 +18,77 @@ import (
...
+
+	deserialized, err = bech32cosmos.ConvertBits(deserialized, 5, 8, false)
+
+	err = cdc.UnmarshalBinary(deserialized, a)
+
+	if err != nil {
+		return err
+	}
+
+	if err != nil {
```
ebuchman, line 93: extra err check ?

pub_key.go
```diff
@@ -76,7 +149,40 @@ func (pubKey PubKeyEd25519) ToCurve25519() *[32]byte {
...
 
 func (pubKey PubKeyEd25519) String() string {
-	return fmt.Sprintf("PubKeyEd25519{%X}", pubKey[:])
+	marshaled, err := cdc.MarshalBinary(pubKey)
+	if err != nil {
+		return err.Error()
+	}
+	conv, err := bech32cosmos.ConvertBits(marshaled, 8, 5, true)
+
+	bech, err := bech32cosmos.Encode("CSMSPUB", conv)
```
ebuchman: lets use constants instead of the magic strings for things we don't give external control over.

Is that right? Addresses will be `CSMSACC` or  `CSMSVAL` but pubkeys will always be `CSMSPUB` ?

pub_key.go
```diff
@@ -76,7 +149,40 @@ func (pubKey PubKeyEd25519) ToCurve25519() *[32]byte {
...
+
+	deserialized, err = bech32cosmos.ConvertBits(deserialized, 5, 8, false)
+
+	err = cdc.UnmarshalBinary(deserialized, pubKey)
+
+	if err != nil {
+		return err
+	}
+
+	if err != nil {
```
ebuchman, line 194: double err

pub_key_test.go
```diff
@@ -10,16 +10,20 @@ import (
...
-		priv: "a96e62ed3955e65be32703f12d87b6b5cf26039ecfa948dc5107a495418e5330",
-		pub:  "02950e1cdfcb133d6024109fd489f734eeb4502418e538c28481f22bce276f248c",
-		addr: "1CKZ9Nx4zgds8tU7nJHotKSDr4a9bYJCa3",
+		priv:       "a96e62ed3955e65be32703f12d87b6b5cf26039ecfa948dc5107a495418e5330",
+		pub:        "02950e1cdfcb133d6024109fd489f734eeb4502418e538c28481f22bce276f248c",
+		addr:       "1CKZ9Nx4zgds8tU7nJHotKSDr4a9bYJCa3",
+		bech32addr: "csmsaddr:cklnrf3g0s4mg25tu6termrk8egltfyme4q7sg3h9myxnv",
+		bech32pub:  "csmspub:ucfk4sszj58peh7tzv7kqfqsnl2gnae5a669qfqcu5uv9pyp7g4uufm0yjxqpj95k8",
 	},
 }
```
ebuchman: we should add some tests for invalid addrs/pubkeys

wire.go
```diff
@@ -16,6 +16,8 @@ func init() {
 }
 
 func RegisterWire(cdc *wire.Codec) {
+	cdc.RegisterConcrete(Address{},
```
ebuchman, line 19: should we be prefix encoding addresses?

ebuchman (commented):
Thanks Zaki!

pub_key.go
zmanian: So I wanted to preserve the same API as `HexBytes` offered and I need an override here because `HexBytes` implements `MarshalJSON`/`UnmarshalJSON`

wire.go
zmanian: I don't know?

zmanian:
I'm having trouble thinking of good failure tests. Open to suggestions

zmanian:
As far as I can tell, this works with go-wire develop.

wire.go
jaekwon: No we shouldn't.  An address is just a []byte, happens to be hex encoded in JSON.  We don't want to register an Address, since it doesn't implement any interface (it's not 1 of anything).

jaekwon:
Commenting on https://github.com/cosmos/cosmos-sdk/issues/263

ebuchman:
We should revive this and get it merged. Looks like it needs a rebase.

Other points:
- the same error is checked twice a couple times (see comments)
- dont amino register addresses

Also, do we need to change the types to structs just so we can memoize the humanreadable ? it might be nicer to just compute the human readable form whenever we need it and leave the type definitions as is. Though we'll probably need `type Address cmn.HexBytes` instead of `type Address = cmn.HexBytes` so we can override the JSON methods

b00f:
If I am not mistaken the bech32cosmos is an error correction and detection algorithm. Regarding to this ,  is it possible to check a given address and make sure it is derived from a public key?

zmanian:
@b00f Because an address is a digest("hash") of a public key, you can't check than an address was generated correctly unless you know the public key value already.

b00f:
As I discussed earlier with @ebuchman it's possible to check address without having public-key by providing checksum at the end of address.
Please check here:
https://github.com/tendermint/abci/pull/245#discussion_r190301914

>In go-crypto (which is a lovely project) it will be helpful if we can add checksum to the address object. This way addresses can be validated by themselves (like Bitcoin addresses), which would avoid much user error by avoiding sends to blackhole addresses ;) :
https://bitcoin.stackexchange.com/questions/32353/how-do-i-check-the-checksum-of-a-bitcoin-address

b00f:
It will be very useful, especially avoiding sending tokens to a mistakenly wrong typed address. Ethereum has been suffering from not having this checksum in its address.

zmanian:
@b00f Yes bech32cosmos adds checksums and will be adopted in the cosmos sdk

zmanian:
See https://github.com/cosmos/cosmos-sdk/pull/1029

b00f:
That will be great. May I know how we can check an address here in go-cypto?

zmanian:
So we decided not to have address verification in go-crypto and instead put it in the sdk

ebuchman:
> So we decided not to have address verification in go-crypto and instead put it in the sdk

I'm wondering if this was the right decision. People not using the SDK will probably want this. I guess the challenge was that its weird to have "COSMOSACCADDR" in the `tendermint/go-crypto` repo ?

PR #76, opened 2018-03-12
WIP: update go-amino (wire)

melekes:


signature_test.go
```diff
@@ -70,17 +70,18 @@ func TestSignatureEncodings(t *testing.T) {
...
 
-		// store as wire
-		bin, err := cdc.MarshalBinary(sig)
+		// store as amino
+		bin, err := cdc.MarshalBinaryBare(sig)
 		require.Nil(t, err, "%+v", err)
 		if tc.sigSize != 0 {
-			assert.Equal(t, tc.sigSize+4, len(bin))
+			// Q: where is 1 byte coming from?
+			assert.Equal(t, tc.sigSize+amino.PrefixBytesLen+1, len(bin))
```
melekes, line 78: Q: where is 1 byte coming from?

jaekwon: Before, go-wire used to not encode array sizes... so `type SignatureEd25519 [64]byte` used to be
just the prefix bytes + the 64 bytes, but now we encode it the same as if it were a slice which happens to have 64 bytes (e.g. is preceded by uvarint(64) which is 1 byte long).

jaekwon (approved):

signature_test.go
melekes: thanks for clarifying this

PR #68, opened 2018-02-15
Remove Ledger

adrianbrink:

PR #85, opened 2018-04-30
Ledger integration

### LINKED ISSUE #93: Panic mitigation
See [this](https://github.com/tendermint/security/issues/6) issue, Crypto-001.

### LINKED ISSUE #99: Package _nano is never build, never used, and doesn't compile
copied over from https://github.com/tendermint/go-crypto/pull/96/files/ecc7cdfdf62350991fbfc092f2e51e4b1ee68091?utf8=%E2%9C%93&diff=unified#r188092482

the package _nano starts with an underscore. `go build` (and with it `go test`) completely ignores this package ...  (see https://golang.org/pkg/go/build/#Context.Import)

The package doesn't build (remove the underscore to reproduce). This is most probably due to: 
https://github.com/golang/go/issues/23672
and a dependency: 
```
go build github.com/ethanfrey/hid: invalid flag in #cgo LDFLAGS: -fconstant-cfstrings
FAIL    github.com/tendermint/go-crypto/nano [build failed]
```
related: https://github.com/ethanfrey/ledger/issues/1#issuecomment-368486919

cwgoes:
Refactored the existing API, although I'm not quite sure if that interface is necessary - did we use to use this in the keybase more directly?

CC @adrianbrink is https://github.com/tendermint/go-crypto/pull/68 a duplicate?

(@Liamsi edited to autom. close related issues on merge):
resolves #99 
resolves #93


cwgoes:
Works now. Use `WITH_LEDGER=1 make test` to run the Ledger testcases (you'll need a physical device plugged in, with the Cosmos app open).

cwgoes:
According to Greg, we can ignore the Jenkins failure.

cwgoes:
- [x] Needs to be updated for some upstream changes

cwgoes:
Can be reviewed.

ValarDragon (approved):
Looks good to me, though I think we may want to test the signing of messages significantly larger than one hash block size.

amino.go
```diff
@@ -19,6 +19,8 @@ func RegisterAmino(cdc *amino.Codec) {
 	cdc.RegisterInterface((*PubKey)(nil), nil)
 	cdc.RegisterConcrete(PubKeyEd25519{},
 		"tendermint/PubKeyEd25519", nil)
+	cdc.RegisterConcrete(PubKeyLedgerEd25519{},
+		"tendermint/PubKeyLedgerEd25519", nil)
```
liamsi: (Maybe stupid) question: Do we actually need to support `Ed25519` here? Isn't this ledger device for end-user wallets? My understanding was we do only use `Secp256k1` for user wallets. Might have misunderstood that though.

amino.go
```diff
@@ -27,6 +29,10 @@ func RegisterAmino(cdc *amino.Codec) {
 		"tendermint/PrivKeyEd25519", nil)
 	cdc.RegisterConcrete(PrivKeySecp256k1{},
 		"tendermint/PrivKeySecp256k1", nil)
+	cdc.RegisterConcrete(PrivKeyLedgerSecp256k1{},
+		"tendermint/PrivKeyLedgerSecp256k1", nil)
+	cdc.RegisterConcrete(PrivKeyLedgerEd25519{},
+		"tendermint/PrivKeyLedgerEd25519", nil)
```
liamsi: same as above

ledger_ed25519.go
```diff
@@ -0,0 +1,156 @@
+package crypto
+
+import (
+	"fmt"
+	"github.com/pkg/errors"
+
+	// "github.com/tendermint/ed25519"
```
liamsi: delete

Gopkg.lock
```diff
@@ -1,17 +1,23 @@
 # This file is autogenerated, do not edit; changes may be undone by the next 'dep ensure'.
 
 
+[[projects]]
+  branch = "master"
+  name = "github.com/brejski/hid"
```
liamsi, line 6: Question: Is this package used anywhere? Can't find it. Is it necessary to include it because of ledger-goclient ?

ledger_ed25519.go
```diff
@@ -0,0 +1,156 @@
...
+	if pk.CachedPubKey == nil {
+		pk.CachedPubKey = pub
+	} else if !pk.CachedPubKey.Equals(pub) {
+		panic("Stored key does not match signing key")
+	}
+	return sig
+}
+
+// PubKey returns the stored PubKey
+// TODO: query the ledger if not there, once it is not volatile
```
liamsi: Is there an issue that keeps track of the open TODOs here? 
Group them together and open issues for them. Ideally assign them to someone. Otherwise, they might be forgotten and always stay todos in the code.

ledger_ed25519.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	return pk.CachedPubKey, err
+}
+
+// forceGetPubKey is like getPubKey but ignores any cached key
+// and ensures we get it from the ledger itself.
+func (pk PrivKeyLedgerEd25519) forceGetPubKey() (key PubKey, err error) {
+	dev, err := getLedger()
+	if err != nil {
+		return key, errors.New(fmt.Sprintf("Cannot connect to Ledger device - error: %v", err))
```
liamsi: `fmt.Errorf` instead of `errors.New(fmt.Sprintf(..`? (here and everywhere else)?

ledger_ed25519.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/pkg/errors"
+
+	// "github.com/tendermint/ed25519"
+	ledger "github.com/zondax/ledger-goclient"
+)
+
+func pubkeyLedgerEd25519(device *ledger.Ledger, path DerivationPath) (pub PubKey, err error) {
+	key, err := device.GetPublicKeyED25519(path)
+	if err != nil {
+		return pub, fmt.Errorf("Error fetching public key: %v", err)
```
liamsi: `pub` will be nil in this case. Make this explicit by writing `return nil, fmt.Errof(...)`. Also, error messages are usually not capitalized (as the error might bubble up and wrapped into another error msg).

ledger_ed25519.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	var p PubKeyLedgerEd25519
+	copy(p[:], key[0:32])
+	return p, err
```
liamsi: `err` should be nil here. Either write this, or:
```golang
var p PubKeyLedgerEd25519
copy(p[:], key[0:32])
pub = p
return
```

ledger_ed25519.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+func signLedgerEd25519(device *ledger.Ledger, path DerivationPath, msg []byte) (sig Signature, err error) {
+	bsig, err := device.SignED25519(path, msg)
+	if err != nil {
+		return sig, err
```
liamsi: just `return`, similarly below

liamsi (changes requested):
LGTM with some minor nitpicks, a few questions

amino.go
cwgoes: Per discussion now removed.

amino.go
cwgoes: Ditto

ledger_ed25519.go
cwgoes: File deleted

ledger_ed25519.go
cwgoes: Updated, lowercased error messages

ledger_ed25519.go
cwgoes: Updated, thanks

ledger_ed25519.go
cwgoes: Updated

ledger_ed25519.go
cwgoes: Updated

Gopkg.lock
cwgoes: Pulled in because of `ledger-goclient`, yes

liamsi (approved):
all comments addressed :-) LGTM

ledger_ed25519.go
cwgoes: Indeed, these were left over from an older version of this code, moved to https://github.com/tendermint/go-crypto/issues/114

adrianbrink:
Awesome work @cwgoes @Liamsi

PR #94, opened 2018-05-09
Use constant-time comparator to compare signatures

### LINKED ISSUE #91: Time signature comparison
See [this](https://github.com/tendermint/security/issues/8) issue, Crypto-003.

liamsi:
resolves #91

### DIFF
--- a/signature.go
+++ b/signature.go
@@ -1,8 +1,9 @@
 package crypto
 
 import (
-	"bytes"
 	"fmt"
+
+	"crypto/subtle"
 
 	. "github.com/tendermint/tmlibs/common"
 )
@@ -41,7 +42,7 @@ func (sig SignatureEd25519) String() string { return fmt.Sprintf("/%X.../", Fing
 
 func (sig SignatureEd25519) Equals(other Signature) bool {
 	if otherEd, ok := other.(SignatureEd25519); ok {
-		return bytes.Equal(sig[:], otherEd[:])
+		return subtle.ConstantTimeCompare(sig[:], otherEd[:]) == 1
 	} else {
 		return false
 	}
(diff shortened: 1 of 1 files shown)


ebuchman:
I’d be interested to understand when non-constant time signature comparison could be exploited. Last time odeke brought this up (https://github.com/tendermint/go-crypto/pull/44), I dug deeper, and even found Adam Langley reverting the use of ConstantTimeCompare because it’s misleading: https://github.com/golang/crypto/commit/c412588e259ef2d488e15762500456cd3b171bad#diff-8183aa2c95dc1385c8f58bf1917fcb0b

veorq:
"Public data" can be misleading, it should be more like "data independent of any secret value" (take for example the classic MAC forgery in Keyczar https://rdist.root.org/2009/05/28/timing-attack-in-google-keyczar-library/).

In the case of EdDSA/ECDSA signatures, an attacker can't leverage non-ct comparators because signature verification works differently—and doesn't use such a comparator anyway. So yeah it looks like a non-ct comparator is okay here.

Naively enforcing ct comparison can be a performance disaster if potentially long strings are compared. But here (and in agl's golang commit) the strings are just 32-byte, so switching back to non-ct will shave off at most 3 cycles in the best case.

ebuchman (approved):

ebuchman:
Copying in more comments from slack to justify this:

----
Well, as long as the message and/or the signature and/or the key itself is not a secret, having constant time verification effectively brings nothing. However, in some _weird_ use cases, a leak might reveal information that should have stayed secret... For example if you were to rely on the so-called "[key-anonymity](https://iacr.org/archive/asiacrypt2001/22480568.pdf)" setting, it might reveal which key you used, and in other such rare cases. 
So it's more some sort of defense in depth recommendation than an actual vulnerability of the current implementation, in my opinion.
I'm surprised Adam removed it from a library which could get used even in such settings.

PR #96, opened 2018-05-09
make sure the passed argument in nano.parseDigest is at least 4 bytes

### LINKED ISSUE #93: Panic mitigation

liamsi:
before accessing its contents 

resolves #93

### DIFF
--- a/_nano/sign.go
+++ b/_nano/sign.go
@@ -40,6 +40,9 @@ func generateSignRequests(payload []byte) [][]byte {
 }
 
 func parseDigest(resp []byte) (key, sig []byte, err error) {
+	if len(resp) < 4 {
+		return nil, nil, errors.Errorf("Incorrect length: %d", len(resp))
+	}
 	if resp[0] != App || resp[1] != Digest {
 		return nil, nil, errors.New("Invalid header")
 	}


_nano/sign.go
```diff
@@ -40,6 +40,9 @@ func generateSignRequests(payload []byte) [][]byte {
 }
 
 func parseDigest(resp []byte) (key, sig []byte, err error) {
+	if len(resp) < 4 {
```
ebuchman, line 43: 4 is len(separator) but looks like we need it to be at least KeyLength because of `resp[KeyLength:]`

liamsi: The problem was that we accessed `resp[0]`, `resp[1]` and `resp[2:]` before checking that these indices actually exist. The `len(resp) < 4` solves this issue. 

Although we access `resp[KeyLength:]` later, we should be fine (no unexpected panic) because we check before we access below:
https://github.com/tendermint/go-crypto/pull/96/files#diff-c053c7317f47fdb9a0ff0ced32c2c4b6R50 
```golang
resp = resp[2:]
if len(resp) != KeyLength+SigLength+len(separator) {
    return nil, nil, errors.Errorf("Incorrect length: %d", len(resp))
}
```

liamsi: > 4 is len(separator) ...

Yes, but we it also the length we'd need to check for the "header" which encodes "App" and "Digest".  

There are still two things that are **weird**, not only about this few lines of code but also about the whole package:
- `resp = resp[2:]` is everything from index 3 onwards ([example](https://play.golang.org/p/aM-RdS3nVSu)). Why do we never do sth.  with `resp[2]` (we only read `resp[0]` and `resp[1]`)?
- the package `_nano` starts with an underscore. `go build` and `go test` completely ignores this package ... (tests are never run as well) (see https://golang.org/pkg/go/build/#Context.Import)

liamsi: See also: https://github.com/tendermint/go-crypto/issues/99

liamsi: This is obsolete with #85

ebuchman: Yep thanks

Issue #102, opened 2018-05-15
Add `.IsValid()` method to pubkey interface


cwgoes:
Ref https://github.com/cosmos/cosmos-sdk/pull/943#discussion_r186583130

PR #104, opened 2018-05-22
Add PubKey.IsValid() and testcases

cwgoes:
Ref https://github.com/tendermint/go-crypto/issues/102

### DIFF
--- a/pub_key.go
+++ b/pub_key.go
@@ -29,6 +29,7 @@ type PubKey interface {
 	Bytes() []byte
 	VerifyBytes(msg []byte, sig Signature) bool
 	Equals(PubKey) bool
+	IsValid() bool
 }
 
 //-------------------------------------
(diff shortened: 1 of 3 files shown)


cwgoes:
Are there more specific tests we can use for checking the validity of a public key (at the moment this just tests nonzero)?

Useful for downstream applications e.g. changing account public keys in the Cosmos SDK (want to prevent the user from setting an invalid public key by accident).

ebuchman:
We could check if the pubkey is on the curve but that would require elliptic curve math so we'd also have to price it right - I think `pubkey.ToCurve25519()` would do it tho. double checking with @zmanian @Liamsi .

Bit concerned about the need for this in the first place though - see https://github.com/cosmos/cosmos-sdk/pull/943#issuecomment-393333841

cwgoes:
Closing as we're reverting https://github.com/cosmos/cosmos-sdk/pull/943 for now.

PR #105, opened 2018-05-22
Bring in merkle and tmhash from tmlibs

ebuchman:
Ref https://github.com/tendermint/tmlibs/pull/210#issuecomment-390995994

### DIFF
--- a/merkle/simple_map_test.go
+++ b/merkle/simple_map_test.go
@@ -1,0 +1,53 @@
+package merkle
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+type strHasher string
+
+func (str strHasher) Hash() []byte {
+	return SimpleHashFromBytes([]byte(str))
+}
+
+func TestSimpleMap(t *testing.T) {
+	{
+		db := NewSimpleMap()
+		db.Set("key1", strHasher("value1"))
+		assert.Equal(t, "acdb4f121bc6f25041eb263ab463f1cd79236a32", fmt.Sprintf("%x", db.Hash()), "Hash didn't match")
+	}
+	{
+		db := NewSimpleMap()
+		db.Set("key1", strHasher("value2"))
+		assert.Equal(t, "b8cbf5adee8c524e14f531da9b49adbbbd66fffa", fmt.Sprintf("%x", db.Hash()), "Hash didn't match")
+	}
+	{
+		db := NewSimpleMap()
+		db.Set("key1", strHasher("value1"))
+		db.Set("key2", strHasher("value2"))
+		assert.Equal(t, "1708aabc85bbe00242d3db8c299516aa54e48c38", fmt.Sprintf("%x", db.Hash()), "Hash didn't match")
+	}
+	{
+		db := NewSimpleMap()
+		db.Set("key2", strHasher("value2")) // NOTE: out of order
+		db.Set("key1", strHasher("value1"))
+		assert.Equal(t, "1708aabc85bbe00242d3db8c299516aa54e48c38", fmt.Sprintf("%x", db.Hash()), "Hash didn't match")
+	}
+	{
+		db := NewSimpleMap()
+		db.Set("key1", strHasher("value1"))
+		db.Set("key2", strHasher("value2"))
+		db.Set("key3", strHasher("value3"))
+		assert.Equal(t, "e728afe72ce351eed6aca65c5f78da19b9a6e214", fmt.Sprintf("%x", db.Hash()), "Hash didn't match")
+	}
+	{
+		db := NewSimpleMap()
+		db.Set("key2", strHasher("value2")) // NOTE: out of order
+		db.Set("key1", strHasher("value1"))
+		db.Set("key3", strHasher("value3"))
+		assert.Equal(t, "e728afe72ce351eed6aca65c5f78da19b9a6e214", fmt.Sprintf("%x", db.Hash()), "Hash didn't match")
+	}
+}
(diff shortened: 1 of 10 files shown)


ebuchman:
Also replaces https://github.com/tendermint/tmlibs/pull/211

liamsi:
referencing discussion on hashing inner & leaf nodes differently; as well as the removed encoding (as both might or might not have security implications): 
https://github.com/tendermint/tmlibs/pull/211#issuecomment-391044446 (and following)

merkle/simple_tree.go
```diff
@@ -0,0 +1,76 @@
...
+*/
+
+package merkle
+
+import (
+	"github.com/tendermint/go-crypto/tmhash"
+)
+
+// Hash(left | right). The basic operation of the Merkle tree.
+func SimpleHashFromTwoHashes(left, right []byte) []byte {
```
liamsi: If the size of left and right is now fixed, can we use the size in the args: 
`SimpleHashFromTwoHashes(left, right [Size]byte)`
see also: https://github.com/tendermint/tmlibs/pull/211#issuecomment-391102658

merkle/simple_tree.go
```diff
@@ -0,0 +1,76 @@
...
+                 /             \
+                *               *
+               / \             / \
+              /   \           /   \
+             /     \         /     \
+            *       *       *       h6
+           / \     / \     / \
+          h0  h1  h2  h3  h4  h5
+
+*/
```
liamsi: Can we add a warning here about pre-image attacks, or, more general a warning on how to use all this safely (in our context, but also in general). See for instance: https://github.com/bitcoin/bitcoin/blob/5961b23898ee7c0af2626c46d5d70e80136578d3/src/consensus/merkle.cpp#L9-L43

liamsi: I can take care of this later too (I would open an issue though referencing relevant comments)

ebuchman: Would love it if you could pick this up! Thanks!

ebuchman:
@Liamsi also worth noting how we use these functions from `github.com/tendermint/types`  - especially that we duplicate the implementation to avoid allocating/populating large slices. 

See PR to update tendermint dependent on this in https://github.com/tendermint/tendermint/pull/1606

liamsi:
Thanks for the pointer! If you like, we can fix the test-vectors here and merge this (so you can reference the development branch there instead: https://github.com/tendermint/tendermint/pull/1606/files#diff-bd247e83efc3c45ae9e8c47233249f18R261). 
I'll keep track and take care of the comments (documenting safety and JP's comments).

merkle/simple_tree.go
liamsi: https://github.com/tendermint/go-crypto/issues/107

merkle/simple_map_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+type strHasher string
+
+func (str strHasher) Hash() []byte {
+	return SimpleHashFromBytes([]byte(str))
```
liamsi: Any reason SimpleHashFromBytes disappeared?
https://github.com/tendermint/tmlibs/blob/50ad19541d61a60bdd6be6f407af3ed7abd6ba44/merkle/simple_tree.go#L70-L74

merkle/simple_tree.go
liamsi: Making this explicit via `left, right [tmhash.Size]byte` makes it necessary to change a lot of places. I propose to keep the encoding (called length prefix here: https://github.com/tendermint/tmlibs/pull/211) and potentially refactor this at a later point.

merkle/simple_map_test.go
ebuchman: meh - trying to simplify the exposed interface and it got replaced with `tmhash.Sum`

liamsi: OK, makes sense

merkle/simple_tree.go
ebuchman: damn ok. my motivation was to simplify the tree description by removing unecessary length prefixing. but i agree that if we do so, we should introduce some better type safety, and doing so will require ALOT of changes ... so maybe its not that important right now ? Just feels like now is a good time to make wanted breaking changes to the data structures :P

ebuchman: Also having length prefixes doesnt solve the 2nd pre-image attack right ?

liamsi: Nope. It just mitigates similar but simpler collisions (e.g., I guess where you just append 2 byte slices together and yield the same root hash as with two leafs ?). Let's  re-add the encoding (which allows for variable sized byte slices to be hashed) and add proper prefixes in a separate issue.

liamsi: > so maybe its not that important right now ? 

it's definitely easier to simply re-introduce the encoding. I've already started adding fixed lengths everywhere in another branch. There seems to be sth wrong though (didn't investigate but I pushed it now: https://github.com/Liamsi/go-crypto/compare/bucky_fixed...Liamsi:fixed_size?expand=1)

ebuchman: sth wrong how ?

ebuchman: also if we go through with https://github.com/tendermint/tmlibs/pull/213 we'll need the length prefix in here.

I guess so long as we use []byte, we should use a length prefix, at least so people don't use it wrong.

liamsi: > sth wrong how ?

TestSimpleProof fails on every index

> I guess so long as we use []byte, we should use a length prefix, at least so people don't use it wrong.

Totally.

ebuchman:
Cool I'm going to squash some commits and then use amino for the encodeByteSlice (didn't want to do that before to avoid an amino dependency in tmlibs)

Issue #107, opened 2018-05-24
Simple Merkle Tree


liamsi:
As with https://github.com/tendermint/go-crypto/pull/105 we moved the merkle and tmhash package from tmlibs to go-crypto. 

This issue makes sure the following issues / comments / discussions do not get lost:

 - [x] merkle: add comment on 2nd pre-image (or other relevant) attacks: https://github.com/tendermint/go-crypto/pull/105#discussion_r190019455
- [ ] merkle: add to package description: when is using this safe and how so? also, differentiate between leaf and interior nodes 
- [x] (if not addressed in #105): are we hashing things of always the same size? ~~if so make this as explicit as possible, if we do not, we need some basic encoding (prefix with len as before~~ len-prefix is back https://github.com/tendermint/tmlibs/pull/211) see: 
https://github.com/tendermint/go-crypto/pull/105#discussion_r190018372 and https://github.com/tendermint/tmlibs/pull/211#issuecomment-391102658

liamsi:
Moved to https://github.com/tendermint/tendermint/issues/1892

PR #108, opened 2018-05-24
revert some changes: testvectors and re-add lenght prefix

liamsi:
~~If we still want to remove the encoding / len prefixing, here is a branch I've played around with different approaches and where I just kept the changes that fix the tests / build issues~~~

I reverted some changes here: 
- use the original test-vectors from tmlib
- (re)add the encoding / length prefix
- fix the tests

Some documentation:
- document public methods and follow golang documentation best practices
- add a warning (and a TODO) about 2nd preimage attacks (see #107) in a separate file (doc.go)

feel free to merge into your branch (if it might save you some time) or ignore and close 

/cc @ebuchman

### DIFF
--- a/merkle/doc.go
+++ b/merkle/doc.go
@@ -1,0 +1,31 @@
+/*
+Package merkle computes a deterministic minimal height Merkle tree hash.
+If the number of items is not a power of two, some leaves
+will be at different levels. Tries to keep both sides of
+the tree the same size, but the left may be one greater.
+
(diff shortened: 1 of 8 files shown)


ebuchman (approved):

ebuchman:
So now I'm not sure what to do because of https://github.com/tendermint/go-crypto/pull/105#discussion_r190622790 ...

ebuchman (approved):
Thank you.

PR #109, opened 2018-05-30
WIP - Adding BLS signatures 

UnitylChaos:
Basic functionality for BLS signatures. Need to make some decisions about interfaces and how we want this structured.

Also the library we are using for the actual pairing implementation (https://github.com/dis2/bls12) imports https://github.com/relic-toolkit/relic/ as a git submodule, which dep is unable to handle, so this PR currently pulls the library repo directly over git in the makefile. We will definitely want to fix that before merge.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,6 +2,15 @@
 
 
 [[projects]]
+  name = "github.com/Project-Arda/bgls"
+  packages = [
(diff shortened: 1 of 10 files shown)


liamsi:
This will be finished after the repos have been consolidated.
Thx @jlandrews!

Issue #103, opened 2018-05-22
ed25519 addresses: remove amino, use SHA256[:20]


ebuchman:
This causes pain in https://github.com/tendermint/go-crypto/issues/77 and is included in https://github.com/tendermint/tendermint/issues/1568

Let's make the PubKeyEd25519.Address() return the first 20 bytes of the hash of the raw 32-byte pubkey, no amino required

See new `tmhash` for the hash function: https://github.com/tendermint/tmlibs/pull/210

liamsi:
resolved via #112

PR #112, opened 2018-05-30
update ed25519 address scheme

ebuchman:
Ref #103

### DIFF
--- a/pub_key.go
+++ b/pub_key.go
@@ -5,11 +5,16 @@ import (
 	"crypto/sha256"
 	"fmt"
 
+	"golang.org/x/crypto/ripemd160"
+
 	secp256k1 "github.com/btcsuite/btcd/btcec"
+
 	"github.com/tendermint/ed25519"
 	"github.com/tendermint/ed25519/extra25519"
+
 	cmn "github.com/tendermint/tmlibs/common"
-	"golang.org/x/crypto/ripemd160"
+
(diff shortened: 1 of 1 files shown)


pub_key.go
```diff
@@ -38,11 +43,9 @@ var _ PubKey = PubKeyEd25519{}
 // Implements PubKeyInner
 type PubKeyEd25519 [32]byte
 
+// Address is the SHA256-20 of the raw pubkey bytes.
 func (pubKey PubKeyEd25519) Address() Address {
-	// append type byte
-	hasher := ripemd160.New()
-	hasher.Write(pubKey.Bytes()) // does not error
-	return Address(hasher.Sum(nil))
+	return Address(tmhash.Sum(pubKey[:]))
```
liamsi, line 48: Just to make sure, as we dropped ed25519 support for wallets, we still need the `Address()` of a `PubKeyEd25519` for other purposes, right?

liamsi (approved):
LGTM. Left one question in the comments.

pub_key.go
ethanfrey: Huh? Just saw this tidbit here...
Is that true? Using only secp? It would be good to add that to BreakingChanges...

ebuchman: will still be supported in go-crypto, we're just not going to have tooling for it for user keys at launch

PR #115, opened 2018-06-03
Move from tmlibs #213

mossid:
Few modifications on merkle. Moved from https://github.com/tendermint/tmlibs/pull/213. Needed for https://github.com/cosmos/cosmos-sdk/pull/1062.

Edit (by @Liamsi), this is needed for: 
- cosmos/cosmos-sdk#823
- cosmos/cosmos-sdk#840
- cosmos/cosmos-sdk#786

### DIFF
--- a/merkle/simple_proof.go
+++ b/merkle/simple_proof.go
@@ -27,7 +27,7 @@ func SimpleProofsFromHashers(items []Hasher) (rootHash []byte, proofs []*SimpleP
 // SimpleProofsFromMap generates proofs from a map. The keys/values of the map will be used as the keys/values
 // in the underlying key-value pairs.
 // The keys are sorted before the proofs are computed.
-func SimpleProofsFromMap(m map[string]Hasher) (rootHash []byte, proofs []*SimpleProof) {
+func SimpleProofsFromMap(m map[string]Hasher) (rootHash []byte, proofs map[string]*SimpleProof) {
 	sm := newSimpleMap()
 	for k, v := range m {
 		sm.Set(k, v)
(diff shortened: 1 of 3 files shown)


merkle/simple_proof.go
```diff
@@ -27,7 +27,7 @@ func SimpleProofsFromHashers(items []Hasher) (rootHash []byte, proofs []*SimpleP
 // SimpleProofsFromMap generates proofs from a map. The keys/values of the map will be used as the keys/values
 // in the underlying key-value pairs.
 // The keys are sorted before the proofs are computed.
-func SimpleProofsFromMap(m map[string]Hasher) (rootHash []byte, proofs []*SimpleProof) {
+func SimpleProofsFromMap(m map[string]Hasher) (rootHash []byte, proofs map[string]*SimpleProof, keys []string) {
```
liamsi, line 30: What is the reason for returning the keys here too? Aren't they included in the newly returned map?

liamsi: Ah OK. You need them sorted in the order as they were included and they are sorted using `.Sort()` again in https://github.com/tendermint/go-crypto/blob/9211fe9dd6cc5ad8a1052739fa674982fc914964/merkle/simple_map.go#L41

liamsi (commented):
This all LGTM! It would be cool to link to the issues that explain why these change is necessary. Then, let's merge it

liamsi (approved):
OK, this is needed for: https://github.com/cosmos/cosmos-sdk/issues/823
https://github.com/cosmos/cosmos-sdk/issues/840
https://github.com/cosmos/cosmos-sdk/issues/786

Issue #116, opened 2018-06-04
Seed phrase has same last word  when create a new key


chengwenxi:
When I create a new key, it will return seed phrase with the same last word every time.
This should be caused by keybase.go#50,

`typ := cryptoAlgoToByte(algo)`
`secret = append([]byte{typ}, secret...)`

secret[0] is invariable, when BytesToWords, because of secret.length = 21 and words.length = 2048(2^11),  the last word index = f(secret.bits[21 * 8-11 * 15]) = f(secret.bits[3]) = f(secret[0].bits[0,2]),is also invariable.

`typ := cryptoAlgoToByte(algo)`
`secret = append(secret, []byte{typ}...)`
will avoid it.

liamsi:
Thanks a lot for raising this issue! We are currently refactoring the code you mentioned and the key-type won’t be encoded in like above anymore. Also, we will only support Secp256k1 keys initially. This makes the whole key-type obsolete.

Related issues/discussions: https://github.com/tendermint/go-crypto/pull/85#discussion_r191945815 and https://github.com/tendermint/go-crypto/issues/89#issuecomment-392864261

chengwenxi:
Yes, it's important to be compatible with cosmos fundraiser keys.

liamsi:
This should be obsolete now!

PR #117, opened 2018-06-05
Unify local and external keys in keybase interface

cwgoes:
Rework `PrivKey`, `Keybase`, and `Info` to work with both locally stored and external keys (offline, HW wallets, etc).

This involves the following API changes:
- Return errors on `privkey.Sign()` and `privkey.PubKey()`, as we may be querying an external datastore
- Add methods to the `Keybase` interface to track Ledger HD ed25519 keys and offline keys, so that signing by name automatically looks up the right key and prompts the user for a password / communicates with the Ledger / prompts the user to sign offline, as appropriate.

### DIFF
--- a/keys/keybase.go
+++ b/keys/keybase.go
@@ -38,7 +38,7 @@ func (kb dbKeybase) Create(name, passphrase string, algo CryptoAlgo) (Info, stri
 	secret := crypto.CRandBytes(16)
 	priv, err := generate(algo, secret)
 	if err != nil {
-		return Info{}, "", err
+		return nil, "", err
 	}
 
 	// encrypt and persist the key
@@ -63,7 +63,7 @@ func (kb dbKeybase) Recover(name, passphrase, seedphrase string) (Info, error) {
 	words := strings.Split(strings.TrimSpace(seedphrase), " ")
 	secret, err := kb.codec.WordsToBytes(words)
 	if err != nil {
-		return Info{}, err
+		return nil, err
 	}
 
 	// secret is comprised of the actual secret with the type
@@ -73,7 +73,7 @@ func (kb dbKeybase) Recover(name, passphrase, seedphrase string) (Info, error) {
 	algo := byteToCryptoAlgo(typ)
 	priv, err := generate(algo, secret)
 	if err != nil {
-		return Info{}, err
+		return nil, err
 	}
 
 	// encrypt and persist key.
@@ -205,18 +205,22 @@ func (kb dbKeybase) Update(name, oldpass, newpass string) error {
 	if err != nil {
 		return err
 	}
-	key, err := unarmorDecryptPrivKey(info.PrivKeyArmor, oldpass)
-	if err != nil {
-		return err
+	switch info.(type) {
+	case localInfo:
+		key, err := unarmorDecryptPrivKey(info.PrivKeyArmor, oldpass)
+		if err != nil {
+			return err
+		}
+		kb.writeKey(key, name, newpass)
+		return nil
+	default:
+		return fmt.Errorf("Locally stored key required")
 	}
-
-	kb.writeKey(key, name, newpass)
-	return nil
 }
 
 func (kb dbKeybase) writePubKey(pub crypto.PubKey, name string) Info {
 	// make Info
-	info := newInfo(name, pub, "")
+	info := newLocalInfo(name, pub, "")
 
 	// write them both
 	kb.db.SetSync(infoKey(name), info.bytes())
@@ -227,10 +231,19 @@ func (kb dbKeybase) writeKey(priv crypto.PrivKey, name, passphrase string) Info 
 	// generate the encrypted privkey
 	privArmor := encryptArmorPrivKey(priv, passphrase)
 	// make Info
-	info := newInfo(name, priv.PubKey(), privArmor)
+	info := newLocalInfo(name, priv.PubKey(), privArmor)
 
 	// write them both
 	kb.db.SetSync(infoKey(name), info.bytes())
+	return info
+}
+
+func (kb dbKeybase) writeLedgerKey(pub crypto.PubKey, path crypto.DerivationPath, name string) Info {
+	// generate the ledger info
+	info := newLedgerInfo(name, pub, path)
+
+	// write the ledger info
+	kb.db.SetSync(ledgerInfoKey(name), info.bytes())
 	return info
 }
 
@@ -248,4 +261,8 @@ func generate(algo CryptoAlgo, secret []byte) (crypto.PrivKey, error) {
 
 func infoKey(name string) []byte {
 	return []byte(fmt.Sprintf("%s.info", name))
+}
+
+func ledgerInfoKey(name string) []byte {
+	return []byte(fmt.Sprintf("ledger.%s.info", name))
 }
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -142,7 +142,8 @@
   packages = [
     "common",
     "db",
-    "log"
+    "log",
+    "test"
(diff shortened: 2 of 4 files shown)


cwgoes:
To-do:

- [x] Add way to track a watch-only key by pubkey
- [x] Correctly switch on `.Sign()` based on keytype
- [x] Add testcases
- [x] Test downstream in Cosmos SDK

cwgoes:
@Liamsi Thoughts on the API changes?

liamsi:
The interface changes look good to me! Totally in favour of sign returning an error. I also like changing Info to an interface. I might change the `Create` to `CreateMneomnicKey` or simply `CreateMnemonic` (as the key is clear because of the context/package name/iface name) in https://github.com/tendermint/go-crypto/pull/118

keys/keybase.go
```diff
@@ -175,23 +227,37 @@ func (kb dbKeybase) ImportPubKey(name string, armor string) (err error) {
...
+		_, err = unarmorDecryptPrivKey(linfo.PrivKeyArmor, passphrase)
+		if err != nil {
+			return err
+		}
+		kb.db.DeleteSync(infoKey(name))
+		return nil
+	case ledgerInfo:
+	case offlineInfo:
+		if passphrase != "yes" {
+			return fmt.Errorf("Enter exactly 'yes' to delete the key")
```
liamsi: nitpick: Error messages shouldn be capitalized. Ideally, run golint before finalizing the PR.

cwgoes: Addressed. Agreed on `golint`, unfortunately this repo is very noncompliant at the moment...

liamsi: Yeah, you can run go-lint only on the files you've modified. That would be great.. Otherwise, don't bother. I'll delete so much code in #116 ... after that nothing is left and it's secure and very compliant ;-)

encode_test.go
```diff
@@ -69,14 +69,16 @@ func TestKeyEncodings(t *testing.T) {
...
+		assert.Nil(t, err)
 		checkAminoBinary(t, sig1, &sig2, -1) // Siganture size changes for Secp anyways.
 		assert.EqualValues(t, sig1, sig2)
 		checkAminoJSON(t, sig1, &sig3, false) // TODO also check Prefix bytes.
 		assert.EqualValues(t, sig1, sig3)
 
 		// Check (de/en)codings of PubKeys.
-		pubKey := tc.privKey.PubKey()
+		pubKey, err := tc.privKey.PubKey()
+		assert.Nil(t, err)
```
liamsi: Can we move this newly introduced asserts to `assert.NoError`? It slightly increases the readability of the tests.

keys/keybase.go
```diff
@@ -110,17 +134,45 @@ func (kb dbKeybase) Sign(name, passphrase string, msg []byte) (sig crypto.Signat
...
+		linfo := info.(ledgerInfo)
+		priv, err = crypto.NewPrivKeyLedgerSecp256k1(linfo.Path)
+		if err != nil {
+			return
+		}
+	case offlineInfo:
+		linfo := info.(offlineInfo)
+		fmt.Printf("Bytes to sign:\n%s", msg)
+		buf := bufio.NewReader(os.Stdin)
+		fmt.Printf("\nEnter encoded signature:\n")
```
liamsi: Be more specific: encoded in what?

keys/keybase.go
```diff
@@ -110,17 +134,45 @@ func (kb dbKeybase) Sign(name, passphrase string, msg []byte) (sig crypto.Signat
...
+		signed, err := buf.ReadString('\n')
+		if err != nil {
+			return nil, nil, err
+		}
+		cdc.MustUnmarshalBinary([]byte(signed), sig)
```
liamsi, line 165: Can you add a comment, that 1) this will block until the user inputs the signature and 2) this assumes that the offline computer is aware of the structure to be signed (TXs?) and properly encodes it in amino. Are we writing the offline-signing software, too? Otherwise, it might be reasonable to assume some other more common encoding. Thoughts?

keys/keybase.go
```diff
@@ -175,23 +227,37 @@ func (kb dbKeybase) ImportPubKey(name string, armor string) (err error) {
...
+		_, err = unarmorDecryptPrivKey(linfo.PrivKeyArmor, passphrase)
+		if err != nil {
+			return err
+		}
+		kb.db.DeleteSync(infoKey(name))
+		return nil
+	case ledgerInfo:
+	case offlineInfo:
+		if passphrase != "yes" {
+			return fmt.Errorf("enter exactly 'yes' to delete the key")
```
liamsi, line 257: Is 'yes' always a user entered value? Otherwise: "call with seedphrase 'yes' to delete the key".

keys/keybase.go
```diff
@@ -205,36 +271,53 @@ func (kb dbKeybase) Update(name, oldpass, newpass string) error {
...
+		return fmt.Errorf("Locally stored key required")
 	}
-
-	kb.writeKey(key, name, newpass)
-	return nil
 }
 
-func (kb dbKeybase) writePubKey(pub crypto.PubKey, name string) Info {
+func (kb dbKeybase) writeLocalKey(priv crypto.PrivKey, name, passphrase string) Info {
+	// generate the encrypted privkey
```
liamsi: Does this actually generate anything? Proposal:
// encrypt private key using passphrase

priv_key.go
```diff
@@ -31,23 +30,19 @@ var _ PrivKey = PrivKeyEd25519{}
 type PrivKeyEd25519 [64]byte
 
 func (privKey PrivKeyEd25519) Bytes() []byte {
-	bz, err := cdc.MarshalBinaryBare(privKey)
-	if err != nil {
-		panic(err)
-	}
-	return bz
+	return cdc.MustMarshalBinaryBare(privKey)
```
liamsi, line 33: Hmm, unrelated to your code but why do we amino encode these bytes?

liamsi (dismissed):
LGTM, left some nitpicks in comments

encode_test.go
cwgoes: All instances changed to `assert.NoError`.

keys/keybase.go
cwgoes: `Amino-encoded signature` (updated), for now at least.

keys/keybase.go
cwgoes: I agree that requiring Amino is suboptimal, let's track that separately though - https://github.com/tendermint/go-crypto/issues/121.

Added a comment about blocking.

keys/keybase.go
cwgoes: In the only downstream consumer I know of, it is user-entered. Leaving for now, if in the future we have other consumers we should probably just change this API anyways.

keys/keybase.go
cwgoes: Agreed, reworded file so "generate" is only used for actually *generating keys*.

priv_key.go
cwgoes: https://github.com/tendermint/go-crypto/issues/121

keys/keybase.go
ValarDragon: Can this support `y` also? (Alot of people are probably in the habit of this)

cwgoes: Most bash commands don't delete keys (even just public keys, as in this case) though. IMO better to require that the user enter exactly 'yes'.

keys/keybase.go
liamsi: Great, thanks! I'll join the discussion in #121 👍

liamsi (approved):
LGTM

PR #118, opened 2018-06-07
keys package: fundraiser compatibility and HD keys (BIP 39 & BIP 32 / BIP 44)

### LINKED ISSUE #57: keys: Recover from seed without persisting to disk
We should be able to enter a seed and use a privkey without that privkey ever being saved to disk

### LINKED ISSUE #89: Word codec should implement BIP39
`keys/words/wordcodec.go` should implement BIP39.

I believe it currently uses an adhoc ECC checksum. Instead it should follow https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki

In particular, it must be compatible with the fundraiser keys, ie. https://github.com/cosmos/fundraiser-lib/blob/master/src/wallet.js

(edit by @Liamsi) resolves #116

### LINKED ISSUE #90: keys: Expose HD tooling
We need to be able to recover private keys from the words written down in the fundraiser, which used HD (bip32). We already have the HD code under `keys/hd` but we don't expose it in any tooling. 

We should have some way to specify an HD path when generating a secp256k1 key (note this only works for secp for now, not ed25519). 

This is specifically to facilitate offline transactions, eg https://github.com/cosmos/cosmos-sdk/issues/966

### LINKED ISSUE #124: Improve error message for the word codec?
Would it be possible to improve the error message [here](https://github.com/tendermint/go-crypto/blob/develop/keys/words/wordcodec.go#L163) in the word codec? I haven't looked at the relevant part of the code to debug what the possible types of errors there are though, but the current error message isn't helpful for debugging what a user did wrong. (Though the cause of the error might be unclear)

For context, I was running:
```
$ democli keys add test --recover
Enter a passphrase for your key:
Repeat the passphrase:
Enter your recovery seed phrase:
apple banana cat dog
ERROR: Invalid data, could not have been generated by this codec
```

liamsi:
resolves #90 
resolves #89
resolves #57

partly resolves #124
closes https://github.com/cosmos/cosmos-sdk/issues/872


armor.go
```diff
@@ -4,23 +4,23 @@ import (
...
+	cmn "github.com/tendermint/tmlibs/common"
 	"golang.org/x/crypto/openpgp/armor"
 )
 
 func EncodeArmor(blockType string, headers map[string]string, data []byte) string {
 	buf := new(bytes.Buffer)
 	w, err := armor.Encode(buf, blockType, headers)
 	if err != nil {
-		PanicSanity("Error encoding ascii armor: " + err.Error())
+		cmn.PanicSanity("Error encoding ascii armor: " + err.Error())
```
xla: AFAIK we want to transition away from panic methods from `common` and use naked panic statements.

keys/hd/fundraiser_test.go
```diff
@@ -0,0 +1,85 @@
...
+		privB, _ := hex.DecodeString(d.Priv)
+		pubB, _ := hex.DecodeString(d.Pub)
+		addrB, _ := hex.DecodeString(d.Addr)
+		seedB, _ := hex.DecodeString(d.Seed)
+		masterB, _ := hex.DecodeString(d.Master)
+
+		seed := bip39.NewSeed(d.Mnemonic, "")
+
+		//fmt.Println("================================")
+		//fmt.Println("ROUND:", i, "MNEMONIC:", d.Mnemonic)
```
xla: Why not always print those and use `t.Log` to have it only presented on verbosre test runs.

keys/hd/hdpath.go
```diff
@@ -0,0 +1,167 @@
...
+func (p BIP44Params) String() string {
+	var changeStr string
+	if p.change {
+		changeStr = "1"
+	} else {
+		changeStr = "0"
+	}
+	// m / purpose' / coin_type' / account' / change / address_index
+	return fmt.Sprintf("%d'/%d'/%d'/%s/%d",
+		p.purpose, p.coinType, p.account, changeStr, p.addressIdx)
```
xla: If we break method calls into multiple lines, can we consistently have the following format:

``` go
	namespace.Method(
		arg1,
		arg2,
		arg3,
	)
```

keys/hd/hdpath.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// ComputeMastersFromSeed returns the master public key, master secret, and chain code in hex.
+func ComputeMastersFromSeed(seed []byte) (secret [32]byte, chainCode [32]byte) {
+	masterSecret := []byte("Bitcoin seed")
+	secret, chainCode = i64(masterSecret, seed)
```
xla, line 84: Is there significance to the named returns, or can we just return the result from `i64` directly?

keys/hd/hdpath.go
```diff
@@ -0,0 +1,167 @@
...
+
+	return
+}
+
+// DerivePrivateKeyForPath derives the private key by following the BIP 32/44 path from privKeyBytes,
+// using the given chainCode.
+func DerivePrivateKeyForPath(privKeyBytes [32]byte, chainCode [32]byte, path string) (derivedKey [32]byte, err error) {
```
xla: Named returns for function bodies of this complexity seem hard to decypher what the state at what point is. It could be clearer with early and explicit returns.

keys/hd/hdpath.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func uint32ToBytes(i uint32) []byte {
+	b := [4]byte{}
+	binary.BigEndian.PutUint32(b[:], i)
+	return b[:]
+}
+
+// i64 returns the two halfs of the SHA512 HMAC of key and data.
+func i64(key []byte, data []byte) (IL [32]byte, IR [32]byte) {
```
xla, line 160: All caps is very unusual, is there a specific reason those can't be `il` and `il` respectively.

keys/hd/hdpath_test.go
```diff
@@ -0,0 +1,73 @@
...
+	// ether
+	priv, _ = DerivePrivateKeyForPath(master, ch, "44'/60'/0'/0/0")
+	fmt.Println(hex.EncodeToString(priv[:]))
+
+	fmt.Println()
+	fmt.Println("keys generated via https://coinomi.com/recovery-phrase-tool.html")
+	fmt.Println()
+
+	seed = bip39.MnemonicToSeed(
+		"advice process birth april short trust crater change bacon monkey medal garment " +
```
xla, line 37: Why not use an unescaped raw string?

keys/keybase.go
```diff
@@ -7,62 +7,111 @@ import (
...
+
+// Language is a language to create the BIP 39 mnemonic in.
+// Currently, only english is supported though.
+// Find a list of all supported languages in the BIP 39 spec (word lists).
+type Language int
+
+const (
+	// English is the default language to create a mnemonic.
+	// It is the only supported language by this package.
+	English Language = iota
```
xla: Would be nice to have the first value at `iota + 1` to prevent accidental zero value passing.

keys/keybase.go
```diff
@@ -7,62 +7,111 @@ import (
...
 // already stored under the same name.
-func (kb dbKeybase) CreateMnemonic(name, passphrase string, algo SignAlgo) (Info, string, error) {
-	// NOTE: secret is SHA256 hashed by secp256k1 and ed25519.
-	// 16 byte secret corresponds to 12 BIP39 words.
-	// XXX: Ledgers use 24 words now - should we ?
-	secret := crypto.CRandBytes(16)
-	priv, err := generate(algo, secret)
+func (kb dbKeybase) CreateMnemonic(name string, language Language, passwd string, algo SigningAlgo) (info Info, mnemonic string, err error) {
+	if language != English {
+		return nil, "", fmt.Errorf("unsupported language: currently only english is supported")
```
xla: Would be nice to have solid error handling, either with sentinal values or types, so that callers get control flow.

keys/keybase.go
```diff
@@ -7,62 +7,111 @@ import (
...
+
+	// default number of words (24):
+	mnemonicS, err := bip39.NewMnemonic(bip39.FreshKey)
 	if err != nil {
-		return nil, "", err
+		return
 	}
+	// TODO(ismail): we have to be careful with the separator in non-ltr languages. Ideally, our package should provide
+	// a helper function for that
+	mnemonic = strings.Join(mnemonicS, " ")
```
xla: The least we can do with the separators is to have them in constants, and ideally later configurable. This is the second occurrence of the literal string and a TODO, looks like a good candidate to fix before merge.

keys/keybase.go
xla: Should we consider prefixing with `Language`.

keys/keys.go
```diff
@@ -1,32 +1,12 @@
...
-	case 0x01:
-		return AlgoEd25519
-	case 0x02:
-		return AlgoSecp256k1
-	default:
-		panic(fmt.Sprintf("Unexpected type byte %X", b))
-	}
-}
+	// Secp256k1 uses the Bitcoin secp256k1 ECDSA parameters.
+	Secp256k1 = SigningAlgo("secp256k1")
```
xla, line 8: Again would prefix.

xla (approved):

Some minor nitpicks. Given my limited understanding from the design doc this looks good to me.

armor.go
liamsi: Thx! done

keys/hd/fundraiser_test.go
liamsi: Sure, why not 👍done

keys/hd/hdpath.go
liamsi: done

keys/keybase.go
liamsi: iota+1 it is 👍 

> Should we consider prefixing with Language.

You mean s/English/LanguageEnglish/ ? Why? Isn't that a bit redundant?

keys/keybase.go
liamsi: There is not much the callers can control here: In these cases (currently) only one possible language is supported (all other will err). This means, if the caller calls this with a different language then engl. he (currently) isn't using the lib correctly (and he can't handle this error case differently then by picking a supported language). Maybe, the sentinel types are still useful as they additionally document the behaviour (additionally to the documentation).

keys/keys.go
liamsi: Same as above.

keys/keybase.go
xla: There is a case to make for either on the caller side it's a bit more clear when reading, yet the type system should inform that already. Could be either way, in general it's good to prefix if the namespace of the package gets a bit more crowded.

keys/bip39/wordcodec.go
```diff
@@ -0,0 +1,62 @@
...
+
+// MnemonicToSeed creates a BIP 39 seed from the passed mnemonic (with an empty BIP 39 password).
+// This method does not validate the mnemonics checksum.
+func MnemonicToSeed(mne string) (seed []byte) {
+	// we do not checksum here...
+	seed = bip39.NewSeed(mne, "")
+	return
+}
+
+// MnemonicToSeedWithErrChecking is completely equivalent to MnemonicToSeed.
```
cwgoes: This comment and the one two lines below seem to contradict each other - by "equivalent" do you mean that it returns the same seed?

keys/hd/hdpath.go
```diff
@@ -0,0 +1,179 @@
...
+		change:     change,
+		addressIdx: addressIdx,
+	}
+}
+
+// NewFundraiserParams creates a BIP 44 parameter object from the params:
+// m / 44' / 118' / account' / 0 / address_index
+// The fixed parameters (purpose', coin_type', and change) are determined by what was used in the fundraiser.
+func NewFundraiserParams(account uint32, addressIdx uint32) *BIP44Params {
+	return &BIP44Params{
```
cwgoes: Might be a bit more idiomatic to call our own `NewParams` here

keys/hd/hdpath.go
```diff
@@ -0,0 +1,179 @@
...
+	secret, chainCode = i64(masterSecret, seed)
+
+	return
+}
+
+// DerivePrivateKeyForPath derives the private key by following the BIP 32/44 path from privKeyBytes,
+// using the given chainCode.
+func DerivePrivateKeyForPath(privKeyBytes [32]byte, chainCode [32]byte, path string) (derivedKey [32]byte, err error) {
+	data := privKeyBytes
+	// TODO(ismail): refactor this out and provide a constructor for BIP44
```
cwgoes: Can we turn this into an issue if it will not be included in this PR?

keys/hd/hdpath.go
```diff
@@ -0,0 +1,179 @@
...
+// If harden is true, the derivation is 'hardened'.
+// It returns the new private key and new chain code.
+func derivePrivateKey(privKeyBytes [32]byte, chainCode [32]byte, index uint32, harden bool) ([32]byte, [32]byte) {
+	var data []byte
+	if harden {
+		index = index | 0x80000000
+		data = append([]byte{byte(0)}, privKeyBytes[:]...)
+	} else {
+		// this can't return an error:
+		pubkey, err := crypto.PrivKeySecp256k1(privKeyBytes).PubKey()
```
cwgoes: Can be rebased on `develop`, which no longer errors on `.PubKey()`

keys/hd/hdpath.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	n := copy(derivedKey[:], data[:])
+	if n != 32 || len(data) != 32 {
+		err = fmt.Errorf("expected a (secp256k1) key of length 32, got length: %v", len(data))
+	}
+	return
+}
+
+// derivePrivateKey derives the private key with index and chainCode.
+// If harden is true, the derivation is 'hardened'.
```
cwgoes: What is the purposing of "hardening" - can we link to the spec?

keys/hd/hdpath_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"encoding/hex"
+	"fmt"
+	"github.com/tendermint/go-crypto/keys/bip39"
+)
+
+func ExampleStringifyPathParams() {
+	path := NewParams(44, 0, 0, false, 0)
+	fmt.Println(path.String())
```
cwgoes, line 11: Why does this print instead of checking the result?

keys/hd/hdpath_test.go
```diff
@@ -0,0 +1,73 @@
...
+	// Output: 44'/0'/0'/0/0
+}
+
+func ExampleSomeBIP32TestVecs() {
```
cwgoes, line 15: Why do these print instead of checking whether or not the outputs are correct in the testcase?

keys/keybase.go
```diff
@@ -81,29 +136,27 @@ func (kb dbKeybase) CreateOffline(name string, pub crypto.PubKey) (Info, error)
...
 
-	// encrypt and persist key.
-	public := kb.writeLocalKey(priv, name, passphrase)
-	return public, nil
+	// if we have a password, use it to encrypt the private key and store it
+	// else store the public key only
+	if passwd != "" {
+		info = kb.writeLocalKey(crypto.PrivKeySecp256k1(derivedPriv), name, passwd)
+	} else {
+		pubk, err := crypto.PrivKeySecp256k1(derivedPriv).PubKey()
```
cwgoes: Error check can be removed with a rebase on develop

random.go
```diff
@@ -82,15 +82,15 @@ func (ri *randInfo) MixEntropy(seedBytes []byte) {
 	hashBytes32 := [32]byte{}
 	copy(hashBytes32[:], hashBytes)
 	ri.seedBytes = xorBytes32(ri.seedBytes, hashBytes32)
-	// Create new cipher.Block
+	// CreateMnemonic new cipher.Block
```
cwgoes: Accidental find replace?

cwgoes (changes requested):
A few minor nits, needs a rebase on develop for the new `PubKey()` API

random.go
liamsi: Good catch!

keys/hd/hdpath_test.go
liamsi: This prints do exactly that: https://blog.golang.org/examples

keys/hd/hdpath_test.go
liamsi: https://blog.golang.org/examples

keys/hd/hdpath.go
liamsi: 👍

keys/bip39/wordcodec.go
liamsi: 👍

keys/hd/hdpath.go
liamsi: done

keys/hd/hdpath.go
liamsi: Added link to the spec which contains further info.

cwgoes (approved):

PR #123, opened 2018-06-11
WIP: Remove SimpleProofsFromMap 

### LINKED ISSUE #122: Delete SimpleProofsFromMap
Following up on a discussion with @mossid about returning the keys in an array additionally to the map (in  https://github.com/tendermint/go-crypto/pull/115 :

Couldn't we pass a list of key-value pairs `KVPairs` as an argument instead of a map in:
https://github.com/tendermint/go-crypto/blob/710efe576acfcc22ead7781cec67a1265cbba0b0/merkle/simple_proof.go#L30

Then we wouldn't need to sort the map. 

Couldn't we then use `SimpleProofsFromHashers` instead then? @mossid 

After discussing with @mossid: Yes, we can delete this. I'll do it before the next release (of go-crypto). (I'm happy to work on the changes in the SDK to take this off your plate @mossid)

liamsi:
resolves #122 

@mossid @cwgoes @sunnya97 does this still work for every use-case of `SimpleProofsFromMap` in the SDK? Do we actually need a `FromMap` method at all? Couldn't we just use the `FromHashers` method instead? 
We could just use a list of KVPairs, extend them to fulfil the Hasher interface and call `SimpleProofsFromHashers` instead. Then we could delete (almost?) everything in [simple_map.go](https://github.com/tendermint/go-crypto/blob/710efe576acfcc22ead7781cec67a1265cbba0b0/merkle/simple_map.go)

### DIFF
--- a/merkle/simple_proof.go
+++ b/merkle/simple_proof.go
@@ -26,26 +26,20 @@ func SimpleProofsFromHashers(items []Hasher) (rootHash []byte, proofs []*SimpleP
 
 // SimpleProofsFromMap generates proofs from a map. The keys/values of the map will be used as the keys/values
 // in the underlying key-value pairs.
-// The keys are sorted before the proofs are computed.
-func SimpleProofsFromMap(m map[string]Hasher) (rootHash []byte, proofs map[string]*SimpleProof, keys []string) {
+// The keys are used in the order given by the kvs list in the argument. In other words the keys are not sorted
+// before the proofs are computed. The proofs for the keys are returned in the same order.
+func SimpleProofsFromMap(kvs []KVPair) (rootHash []byte, proofs []*SimpleProof) {
 	sm := newSimpleMap()
-	for k, v := range m {
-		sm.Set(k, v)
+	for _, kv := range kvs {
+		sm.Set(string(kv.Key), kv)
 	}
-	sm.Sort()
-	kvs := sm.kvs
+
 	kvsH := make([]Hasher, 0, len(kvs))
 	for _, kvp := range kvs {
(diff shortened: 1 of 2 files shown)


cwgoes:
I don't think there *are* any uses of `SimpleProofsFromMap` in the SDK - @mossid do you use it on your branch?

merkle/simple_proof.go
```diff
@@ -26,26 +26,20 @@ func SimpleProofsFromHashers(items []Hasher) (rootHash []byte, proofs []*SimpleP
...
+// The keys are used in the order given by the kvs list in the argument. In other words the keys are not sorted
+// before the proofs are computed. The proofs for the keys are returned in the same order.
+func SimpleProofsFromMap(kvs []KVPair) (rootHash []byte, proofs []*SimpleProof) {
 	sm := newSimpleMap()
-	for k, v := range m {
-		sm.Set(k, v)
+	for _, kv := range kvs {
+		sm.Set(string(kv.Key), kv)
 	}
-	sm.Sort()
```
mossid, line 35: Shouldn't we add `kvs.Sort()` here instead of making the user of this function handle it?

mossid (changes requested):
LGTM

merkle/simple_proof.go
liamsi: That's a very good idea. I'm in favour of deleting `SimpleProofsFromMap` and using `SimpleProofsFromHashers` instead as discussed in the issue related to this PR #122. Let me know what you think.

liamsi:
This will be worked on after go-crypto was merged into tendermint. Closing for now.

Issue #129, opened 2018-06-12
Proposal: minor API simplifications


liamsi:
- remove error from privKey.PublicKey(), if one already has the priv. key in hand the pub-key should be derivable from that without erroring: resolved via #131 
- remove the `Get*` in front of the `Info` interface methods: "it's neither idiomatic nor necessary to put Get into the getter's name." https://golang.org/doc/effective_go.html?#Getters

cwgoes:
Agreed on both counts. Need to make sure we store and cache the public key when creating a remote private key (e.g. on the Ledger device) to fulfill this API.

liamsi:
- error has been removed in #131 
- discussion regarding the getters should happen in cosmos-sdk

cwgoes:
Moved to https://github.com/cosmos/cosmos-sdk/issues/1518.

PR #131, opened 2018-06-12
Remove error from priv.PubKey()

cwgoes:
Ref https://github.com/tendermint/go-crypto/issues/129


liamsi (approved):
Thanks for taking care of this! 
LGTM!

PR #134, opened 2018-06-14
circleci 2.0

liamsi:
- switching to circleci 2.0: minimalistic config (circleci 1.0 configs won't be supported soon)
image


xla (approved):

PR #135, opened 2018-06-15
Switch xchachapoly to hkdfchachapoly

ValarDragon:


hkdfchacha20poly1305/hkdfchachapoly.go
```diff
@@ -0,0 +1,95 @@
+package xchacha20poly1305
```
zmanian: change the package name

hkdfchacha20poly1305/hkdfchachapoly_test.go
```diff
@@ -1,17 +1,67 @@
...
+	"bytes"
+	cr "crypto/rand"
+	mr "math/rand"
+	"testing"
 )
 
-func toHex(bits []byte) string {
-	return hex.EncodeToString(bits)
-}
+// Copied from https://github.com/golang/crypto/blob/master/chacha20poly1305/chacha20poly1305_test.go#L69
```
ValarDragon: Is there some official way I need to document this?

ValarDragon:
Does this scheme need to be documented anywhere? And if so where? 

What we've done is switch the use of HChaCha20 inside of XChaCha20 to a use of HKDF. XChaCha20 does some stuff with shuffling which part of the nonce goes where. What we've done instead is put the whole inputted nonce (From Seal/ Open) into hkdf, and we read an additional 12 bytes out of the hkdf, to use a nonce for the ChaCha20 instance.

hkdfchacha20poly1305/hkdfchachapoly_test.go
liamsi: We'll have to make sure we follow the requirements of this license and if we can mix with what we use etc:
https://github.com/golang/crypto/blob/a8fb68e7206f8c78be19b432c58eb52a6aa34462/chacha20poly1305/chacha20poly1305_test.go#L1-L3

hkdfchacha20poly1305/hkdfchachapoly.go
```diff
@@ -0,0 +1,87 @@
...
+	"crypto/sha256"
+	"errors"
+	"fmt"
+	"io"
+
+	"golang.org/x/crypto/chacha20poly1305"
+	"golang.org/x/crypto/hkdf"
+)
+
+var sigma = [4]uint32{0x61707865, 0x3320646e, 0x79622d32, 0x6b206574}
```
liamsi: Add a comment what this sigma means and where the hex nums come from.

ValarDragon: Woops forgot to delete those, thanks!

hkdfchacha20poly1305/hkdfchachapoly.go
```diff
@@ -0,0 +1,87 @@
+package hkdfchacha20poly1305
```
liamsi, line 5: Can we a have a high level overview here? It will create a neat godoc entry and helps the user understand what is going on below. Just a few sentences.

hkdfchacha20poly1305/hkdfchachapoly.go
```diff
@@ -0,0 +1,87 @@
...
+
+	var subKey [KeySize]byte
+	var chachaNonce [chacha20poly1305.NonceSize]byte
+
+	hash := sha256.New
+	hkdf := hkdf.New(hash, c.key[:], nonce, []byte(HkdfInfo))
+	io.ReadFull(hkdf, subKey[:])
+	io.ReadFull(hkdf, chachaNonce[:])
+
+	chacha20poly1305, _ := chacha20poly1305.New(subKey[:])
```
liamsi: Usually bad practice to ignore an error. Add a comment why this can't err here or handle the error.

hkdfchacha20poly1305/hkdfchachapoly.go
```diff
@@ -0,0 +1,87 @@
...
+
+	return chacha20poly1305.Seal(dst, chachaNonce[:], plaintext, additionalData)
+}
+
+func (c *hkdfchacha20poly1305) Open(dst, nonce, ciphertext, additionalData []byte) ([]byte, error) {
+	if len(nonce) != NonceSize {
+		return nil, fmt.Errorf("hkdfchacha20poly1305: bad nonce length passed to Open")
```
liamsi: Nitpick: If you don't need a formatting directive, you can use `errors.New` instead (as you did above).

hkdfchacha20poly1305/hkdfchachapoly.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	if uint64(len(ciphertext)) > (1<<38)-48 {
+		return nil, fmt.Errorf("hkdfchacha20poly1305: ciphertext too large")
```
liamsi: same as above

hkdfchacha20poly1305/hkdfchachapoly.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	var subKey [KeySize]byte
+	var chachaNonce [chacha20poly1305.NonceSize]byte
+
+	hash := sha256.New
+	hkdf := hkdf.New(hash, c.key[:], nonce, []byte(HkdfInfo))
+	io.ReadFull(hkdf, subKey[:])
+	io.ReadFull(hkdf, chachaNonce[:])
```
liamsi: These also can err ... Any good reason to ignore those?

hkdfchacha20poly1305/hkdfchachapoly_test.go
```diff
@@ -0,0 +1,67 @@
+package hkdfchacha20poly1305
+
+import (
+	"bytes"
+	cr "crypto/rand"
+	mr "math/rand"
+	"testing"
+)
+
+// Copied from https://github.com/golang/crypto/blob/master/chacha20poly1305/chacha20poly1305_test.go#L69
```
liamsi: See comment about copy-right above.

hkdfchacha20poly1305/hkdfchachapoly.go
```diff
@@ -0,0 +1,87 @@
...
+	copy(ret.key[:], key)
+	return ret, nil
+
+}
+func (c *hkdfchacha20poly1305) NonceSize() int {
+	return NonceSize
+}
+
+func (c *hkdfchacha20poly1305) Overhead() int {
+	return 16
```
liamsi: Should we make this (16) a public constant instead?

hkdfchacha20poly1305/hkdfchachapoly_test.go
liamsi: https://github.com/golang/crypto/blob/master/LICENSE

hkdfchacha20poly1305/hkdfchachapoly.go
liamsi: also, maybe provide the caller with a max. cipher text size too.

hkdfchacha20poly1305/hkdfchachapoly.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func (c *hkdfchacha20poly1305) Seal(dst, nonce, plaintext, additionalData []byte) []byte {
+	if len(nonce) != NonceSize {
+		panic("hkdfchacha20poly1305: bad nonce length passed to Seal")
+	}
+
+	if uint64(len(plaintext)) > (1<<38)-64 {
+		panic("hkdfchacha20poly1305: plaintext too large")
```
liamsi: suggestion: provide the caller with a max. plaintext size?

hkdfchacha20poly1305/hkdfchachapoly.go
```diff
@@ -0,0 +1,87 @@
...
+	return NonceSize
+}
+
+func (c *hkdfchacha20poly1305) Overhead() int {
+	return 16
+}
+
+func (c *hkdfchacha20poly1305) Seal(dst, nonce, plaintext, additionalData []byte) []byte {
+	if len(nonce) != NonceSize {
+		panic("hkdfchacha20poly1305: bad nonce length passed to Seal")
```
liamsi, line 57: any reason this isn't just an error and panics instead?

hkdfchacha20poly1305/hkdfchachapoly.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	var subKey [KeySize]byte
+	var chachaNonce [chacha20poly1305.NonceSize]byte
+
+	hash := sha256.New
+	hkdf := hkdf.New(hash, c.key[:], nonce, []byte(HkdfInfo))
+	io.ReadFull(hkdf, subKey[:])
+	io.ReadFull(hkdf, chachaNonce[:])
+
+	chacha20poly1305, _ := chacha20poly1305.New(subKey[:])
```
liamsi: Add a comment why you can ignore this err.

hkdfchacha20poly1305/hkdfchachapoly.go
```diff
@@ -0,0 +1,87 @@
...
+		panic("hkdfchacha20poly1305: plaintext too large")
+	}
+
+	var subKey [KeySize]byte
+	var chachaNonce [chacha20poly1305.NonceSize]byte
+
+	hash := sha256.New
+	hkdf := hkdf.New(hash, c.key[:], nonce, []byte(HkdfInfo))
+	io.ReadFull(hkdf, subKey[:])
+	io.ReadFull(hkdf, chachaNonce[:])
```
liamsi: Add a comment why this can't err or handle the err.

liamsi (changes requested):
Thanks @ValarDragon 👍 I left a few comments. PTAL.

Looking good in general!

hkdfchacha20poly1305/hkdfchachapoly.go
ValarDragon: We have to adhere to the AEAD API

ValarDragon:
Thanks for the comments! The latest commit addresses them all (to my knowledge)

hkdfchacha20poly1305/hkdfchachapoly.go
liamsi: Of course. Makes sense!

hkdfchacha20poly1305/hkdfchachapoly.go
```diff
@@ -0,0 +1,106 @@
+package hkdfchacha20poly1305
+
+// This package creates an AEAD using hkdf, chacha20, and poly1305
+// When sealing and opening, the hkdf is used to obtain the nonce and subkey for
+// chacha20. Other than the change for the how the subkey and nonce for chacha
+// are obtained, this is the same as chacha20poly1305
```
liamsi: Can you move this above the package? Ideally, have a look how godoc renders before and after the change.
(run `godoc -http=:6060` and open locally in your browser)
image

hkdfchacha20poly1305/hkdfchachapoly_test.go
```diff
@@ -0,0 +1,139 @@
...
+		"fc59773c2c864ee3b4cc971876b3c7bed4",
+		"caec14e3a9a52ce1a2682c6737defa4752",
+		"0b89511ffe490d2049d6950494ee51f919",
+		"7de854ea71f43ca35167a07566c769083d",
+		"cd477327f4ea4765c71e311c5fec1edbfb"}
+
+	for i := 0; i < 10; i++ {
+		ct, _ := hex.DecodeString(cts[i])
+
+		byte_arr := []byte{byte(i)}
```
liamsi: underscores are not very common in go code `byte_arr`

hkdfchacha20poly1305/hkdfchachapoly.go
```diff
@@ -0,0 +1,106 @@
...
+		panic("hkdfchacha20poly1305: bad nonce length passed to Seal")
+	}
+
+	if uint64(len(plaintext)) > MaxPlaintextSize {
+		panic("hkdfchacha20poly1305: plaintext too large")
+	}
+
+	subKey, chachaNonce := getSubkeyAndChachaNonceFromHkdf(&c.key, &nonce)
+
+	chacha20poly1305, err := chacha20poly1305.New(subKey[:])
```
liamsi: can we call this var `aead` instead? (then, it wouldn't be the same as the package name, which is slightly confusing and shorter :-) )

liamsi (changes requested):

liamsi (approved):
Sweet! Thanks @ValarDragon 👍

PR #139, opened 2018-06-15
fix circleci 2.0 config

liamsi:


xla (approved):

PR #142, opened 2018-06-20
delete keys package

liamsi:
- it will be moved to cosmos-sdk
- code needs to be moved to comos-sdk, ideally while preserving the history of the keys package


cwgoes (approved):
utACK

PR #144, opened 2018-07-01
Release/v0.10.0

ebuchman:
When looking into the version of go-crypto that got merged into Tendermint to try and accurately document the breaking changes, I noticed a few things:

- Changelog for v0.8.0 on master was missing some entries
- Changelog for v0.8.0 was missing altogether on develop
- develop was `37 commits ahead, 6 commits behind master`

It looks like develop is what got merged into tendermint, but it has more changes that aren't captured on master or in a release, so we can't actually say "version X of go-crypto was merged into tendermint". This is somewhat unfortunate.

As for the commits on master not on develop, I'm not exactly sure what happened, but some seem to be in Tendermint and some don't - in particular the recent `This is a combination of X commits.` commits are missing.

I have done the following:

- Update the changelog for v0.8.0 on master
- Merge master back into develop (there were some conflicts...)
- Prepare release v0.10.0, the final release for go-crypto

I'm not sure if there is something here that is missing in the tendermint/crypto or in the `keys` that got moved to SDK.

I'm mostly just concerned with accurately documenting the breaking changes in go-crypto so that when we release Tendermint with crypto moved in, we can say exactly what happened.


liamsi:
> It looks like develop is what got merged into tendermint, but it has more changes that aren't captured on master or in a release, so we can't actually say "version X of go-crypto was merged into tendermint". This is somewhat unfortunate.

Yes, develop was merged into tendermint. We thought, we might "retire" go-crypto by making it a private repo as as soon as the merge is through. Hence we took the develop branch and did not create a release before (cc @cwgoes)

> I'm not sure if there is something here that is missing in the tendermint/crypto or in the keys that got moved to SDK.

We deleted the entire keys package here (7ebc7c08e7de58540762d6ab2fc688677bef80c9 and a589e0eabd113cb6f788e8f24b75870c11816937). It is not needed in tendermint. Merging into cosmos-sdk happend exactly like into to tendermint (from develop) but with reverting these two commits (and deleting everything else).

